PR #22, opened 2023-02-14
Add configuration plugin type

g-marconet:
Add a configuration plugin type. Tweak the repo health plugin to work with the configuration plugin type (and to use plugin configs properly)

## Description
- Adds a configuration plugin type
  - I grepped for "static_analysis" and "inventory" and did my best to figure out where it made sense to use "configuration" as well
  - Inventory and configuration are similar, so I tended to follow inventory more closely
- Repo health plugin now uses plugin configurations (like those used by checkov).

## Motivation and Context
A configuration plugin type is necessary for the repo health plugin, as it does not fall neatly into any of the other plugin categories.

## How Has This Been Tested?
- [x] Manually setting up artemis and running scans with `repo_health`
- [x] Added test for the configuration plugin type

## Types of changes

## Checklist


backend/engine/plugins/repo_health/main.py
```diff
@@ -74,6 +75,8 @@ def are_results_passing(results):
 
 def print_and_exit(output, exit_code=0):
     print(json.dumps(output))
+    # TODO remove log
+    log.info(json.dumps(output))
```
davakos, line 79: should this be removed before PR?

backend/lambdas/generators/json_report/json_report/report.py
```diff
@@ -39,12 +41,22 @@ def get_report(scan_id, params=None):
...
+        config_results = get_configuration(scan)
+
     errors.update(vuln_results.errors)
     errors.update(secret_results.errors)
     errors.update(sa_results.errors)
     errors.update(inv_results.errors)
+    errors.update(config_results.errors)
 
-    success = vuln_results.success and secret_results.success and sa_results.success and inv_results.success
+    success = (
```
davakos, line 53: this is one area where I would think that config plugin results may potentially differ from inventory plugins - that we may want failed config checks to contribute to the overall pass/fail `success` of the scan results. if any vuln or static analysis findings are identified (even if lower severity), it still accounts for a "failing" (`false`) `success`, unless a user is specifically filtering for certain criteria, e.g., `severity=critical&severity=high`. 

I would think we would want to replicate that with config results.. if there are no filters, then any test failure contributes to a failed overall result status. **this looks like what you are doing here in this code.**

we may want to discuss offline whether it makes sense to also provide an option to filter-out certain config checks from a failing status result.

backend/lambdas/generators/json_report/tests/test_generate_report.py
```diff
@@ -410,6 +432,28 @@ def test_get_inventory_report_for_technology_discovery(self):
...
+            },
+            PluginErrors(),
+            True,
+            {"repo_health": 1},
+        )
+
+        mock_scan = unittest.mock.MagicMock(side_effect=Scan())
+        mock_scan.pluginresult_set.filter.return_value = [TEST_REPO_HEALTH]
+        configuration = get_configuration(mock_scan)
+        self.assertEqual(expected_configuration, configuration)
```
davakos, line 455: should there also be a test that checks a failing config result and that it sets scan `success` = `false`

backend/engine/plugins/repo_health/main.py
```diff
@@ -13,32 +15,32 @@ def main():
     output = {
         "success": False,
         "truncated": False,
-        "details": [],
+        "details": {},
```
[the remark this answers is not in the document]
g-marconet: Yes, but it shouldn't be breaking. This PR is merging into the feature branch for repo health, not main.

This is how `inventory` plugins handle details, so I figured consistency there would be beneficial.

backend/engine/plugins/repo_health/main.py
g-marconet: Good catch

backend/lambdas/generators/json_report/tests/test_generate_report.py
g-marconet: I've added the test here: backend/engine/tests/test_plugin_repo_health.py

backend/engine/plugins/repo_health/main.py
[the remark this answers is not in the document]
pizen: But does it make sense to be a dictionary? Is it serving the same purpose as with `inventory` plugins?

backend/lambdas/generators/json_report/tests/test_generate_report.py
```diff
@@ -225,6 +226,27 @@
     end_time=datetime(year=2020, month=2, day=19, hour=15, minute=1, second=55, tzinfo=timezone.utc),
 )
 
+TEST_REPO_HEALTH = PluginResult(
+    plugin_name="Repo Health",
+    plugin_type="configuration",
+    success=True,
+    details={
+        "repo_health": [
```
pizen, line 234: What benefit is having the plugin name be a key in the plugin details bringing?

pizen (changes requested):

backend/engine/plugins/repo_health/main.py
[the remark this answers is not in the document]
g-marconet: I'm not sure. What is the rational behind it being a dictionary in Inventory?

From what I can tell, the same thing applies to both inventory and configuration. Inventory has things broken up by the type of inventory (at least that's what it seems like by the technology discovery plugin). To me, it makes sense that Configuration would break things up by the type of configuration check (so, for example, if we had an AWS health check that would be in its own key).

I'm open to doing it however, though

backend/lambdas/generators/json_report/tests/test_generate_report.py
g-marconet: Bringing everything together:

> But does it make sense to be a dictionary? Is it serving the same purpose as with inventory plugins?

I'm not sure what the intent was with `inventory` (the docs are kinda sparse). I interpreted it as a way to segment things within the inventory type. With that interpretation, it makes sense to me to have the same thing for `configuration`, as handling `Github Repo Health` and a hypothetical  `AWS Configuration` plugin would be completely different.

I'm not attached to doing it this way though. What was the intent of the dictionary with the `inventory` type?

backend/lambdas/api/repo/repo/util/const.py
```diff
@@ -46,6 +46,7 @@
         "checkov": None,
     },
     "inventory": {"technology_discovery": None, "base_images": None},
+    "configuration": {"repo_health": None},
```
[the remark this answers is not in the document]
g-marconet: This will be Github-specific, good catch.

backend/engine/plugins/github_repo_health/cli/src/rules/helpers.py
```diff
@@ -25,16 +37,22 @@ def evaluate_array_config(config, eval_fn):
...
-        "type": check.identifier,
+        "id": id,
         "name": name,
         "description": description,
         "pass": passing,
     }
 
+    if severity:
+        # Add severity as an optional configuration
+        result["severity"] = severity
```
pizen, line 54: If the severity is an empty string (a valid severity level) this will evaluate to false and not add the severity to the result. `if severity is not None` might be better.

backend/engine/plugins/github_repo_health/cli/src/rules/helpers.py
```diff
@@ -25,16 +37,22 @@ def evaluate_array_config(config, eval_fn):
 
 
 def add_metadata(passing, check, config={}, error_message=None):
+    id = config.get("id") or check.identifier
     name = config.get("name") or check.name
     description = config.get("description") or check.description
+    severity = config.get("severity")
```
pizen, line 43: Should this have a default value?

backend/engine/plugins/github_repo_health/main.py
```diff
@@ -0,0 +1,127 @@
...
+        },
+        {
+            "type": "repo_security_alerts",
+            "id": "github_repo_security_alerts",
+        },
+        # Refer to engine/plugins/github_repo_health/cli/src/rules for other rules
+    ],
+}
+
+log = utils.setup_logging(PLUGIN_NAME)
```
pizen, line 46: Let's use the `Logger` from `artemislib` and start phasing out this logging function.

pizen (changes requested):

backend/engine/plugins/github_repo_health/cli/src/rules/helpers.py
g-marconet: Good catch. I don't think it will matter in the context of Artemis, since no severity will be made into empty string there anyway, but it'd make sense for `"severity": ""` in the config to explicitly set the severity.

backend/engine/plugins/github_repo_health/cli/src/rules/helpers.py
g-marconet: I don't think so, unless we explicitly want to tell it to be `None`.

The intent is for `severity` to only show up in the output of repo health if it's specified in the config

pizen (approved):

PR #23, opened 2023-02-24
Add plugin for gathering secrets from GHAS Secret Scanning

pizen:
## Description
This is a plugin to retrieve and report secrets identified by the GitHub Advanced Security Secret Scanning. As such, it only works against repositories that are in GitHub and have Secret Scanning enabled. To reduce friction, this plugin is only enabled when the feature is enabled in the Artemis environment, similar to how Veracode, Snyk, and Aqua are managed.

If the feature is enabled and the repository is in GitHub but the Secret Scanning functionality is not enabled on the repository the plugin runs but returns no results. Similarly, if the feature is enabled but the repository is not in GitHub the plugin runs but returns no results.

In order to function successfully, the GitHub app or PAT used by Artemis to access GitHub must have sufficient permissions to read the secret scanning alerts: https://docs.github.com/en/rest/overview/permissions-required-for-github-apps#secret-scanning-alerts

## Motivation and Context
We want to be able to use GHAS Secret Scanning as an additional source of scan data.

## How Has This Been Tested?
- Local testing, both running the plugin directly and in a benchtop deployment of Artemis, against repositories with and without GHAS Secret Scanning.
- Unit testing

## Types of changes
- [X] New feature (non-breaking change which adds functionality)

## Checklist
- [X] I have added tests to cover my changes.

## Pic
Embed something funny here


backend/engine/tests/test_ghas_secrets.py
```diff
@@ -0,0 +1,19 @@
...
+
+class TestGHASSecrets(unittest.TestCase):
+    def test_normalize_secret_type(self):
+        test_cases = [
+            ("aws_access_key_id", "aws"),
+            ("not_aws_key", "not_aws_key"),
+            ("google_api_key", "google"),
+            ("github_ssh_private_key", "ssh"),
+            ("slack_incoming_webhook_url", "slack"),
+            ("slack_workflow_webhook_url", "slack"),
```
davakos, line 14: might want to test the fall-through case too

davakos (commented):
1. Are any updates needed in `artemis-scan.sh` for this new plugin? I assume just having the GHAS results included if the feature is available+enabled is fine for these:
   * https://github.com/WarnerMedia/artemis/blob/main/backend/ci-tools/shell/artemis-scan.sh#L14
   * https://github.com/WarnerMedia/artemis/blob/main/backend/ci-tools/shell/artemis-scan.sh#L101
2. Do you think we should include a `README.md` for this plugin indicating that unlike other Artemis plugins, that run a tool against checked-out code, this plugin uses the GH API to query secret scanning alerts for the repo (a feature of GHAS), and then filters these alerts to only ones matching the scan branch (location)?

backend/engine/tests/test_ghas_secrets.py
pizen: I did. The second test case (line 10) is the fall-through case: `("not_aws_key", "not_aws_key")`

pizen:
@davakos 
> 1. Are any updates needed in `artemis-scan.sh` for this new plugin? I assume just having the GHAS results included if the feature is available+enabled is fine for these:
>    
>    * https://github.com/WarnerMedia/artemis/blob/main/backend/ci-tools/shell/artemis-scan.sh#L14
>    * https://github.com/WarnerMedia/artemis/blob/main/backend/ci-tools/shell/artemis-scan.sh#L101

I think it's fine. For most repos it will add a little overhead of running the plugin and making 1 API call but right now I think that's an acceptable tradeoff for adding in the GHAS coverage where it is enabled.

> 2. Do you think we should include a `README.md` for this plugin indicating that unlike other Artemis plugins, that run a tool against checked-out code, this plugin uses the GH API to query secret scanning alerts for the repo (a feature of GHAS), and then filters these alerts to only ones matching the scan branch (location)?

I guess we could.

backend/engine/plugins/ghas_secrets/main.py
```diff
@@ -0,0 +1,230 @@
...
+    from artemislib.aws import AWSConnect  # pylint: disable=import-outside-toplevel
+
+    aws_connect = AWSConnect()
+    secret = aws_connect.get_secret(f"{APPLICATION}/{service_secret}")
+    if secret:
+        return secret.get("key")
+    return None
+
+
+def _github_api_get(gh: GITHUB_INFO, path: str = None, query: dict = None, paged=False):
```
g-marconet, line 176: Is there a way this can be factored out so that it is usable by other plugins (for example, making it a part of `GithubAPI` in `artemislib`)?

At minimum, I think the header-setting stuff would be good factored out into a common library. The paging stuff would also be useful

davakos (dismissed):

backend/engine/plugins/ghas_secrets/main.py
pizen: Updated

g-marconet (dismissed):
LGTM

davakos (dismissed):

davakos (approved):

PR #40, opened 2023-03-24
Make plugin event processing environment-configurable

pizen:
## Description
- Add support in the engine for env vars that control whether plugin events are processed
- Update TF to plumb configuration of these env vars down to the engine EC2 template
- Update local env script to set the variables

## Motivation and Context
While the engine can process plugin events for secrets, inventory, and configuration plugins not every deployment is going to have need for them. Rather than generating the events and letting them get dropped at the event dispatch lambda, which might generate log messages depending on the event type, make it so each type can be turned on or off. This way only the events that are used will be generated.

## How Has This Been Tested?
- Verified in a test environment that TF applies correctly
- Verified in a test environment that events are only generated for the enabled categories

## Types of changes
- [X] New feature (non-breaking change which adds functionality)

## Checklist
- [X] My code follows conforms to the coding standards.
- [X] I have updated the documentation accordingly.
- [X] All new and existing tests passed.

## Pic
Embed something funny here


g-marconet (commented):
Looks good, but I'm wondering if there would there be any value in doing this generically? For example, with a `DISABLED_EVENTS_TYPE` env var or something that is a list of types that we then exclude.

It'd simplify things a little if we add additional plugin types down the line, but also allows us to disable the other plugin types if a use-case ever arises where we would want to.

pizen:
@g-marconet For this sort of feature control I prefer to use discrete boolean flags.

g-marconet (approved):

PR #50, opened 2023-05-10
New Feature: Email Aliases for Authentication

jlegarreta:
## Description

This feature handles a situation where users' email addresses have changed in a graceful manner. Rather than go into the database and manually change email addresses as they are updated, risking downtime for some users, a variable can be defined with a mapping of the old email format to the new email format. When a user logs in with their new email, the existing user with the old email will be found and automatically updated with the new email.

The `email_domain_aliases` variable is defined in the analyzer module section of a given terraform environment `main.tf` file.

A simple example of the variable assignment would be as follows:
```
email_domain_aliases = [
  {
    new_domain           = "newcompany.com",
    old_domains          = ["oldcompany.com"]
  }
]
```

The authorizer lambda function will read this data, and when a user logs in with an email that is not found in the database, it will check to see if there is an alias. If there is an alias, then it will find the user with the old domain, update the email to the new one, and log the user in.

A more complex example would be as follows:
```
...
      {
        new_domain           = "company.com",
        old_domains          = ["company.com"],
        email_transformation = { new_email_regex = "[.]", old_email_expr = "_" }
      }
...
```

The `email_transformation` variable allows one to transform the local part of the email address as needed. For example, the above handles a situation in which a company maintains the same domain, but changes the local part convention from `first_lastname@company.com` to `first.lastname@company.com`. The `new_email_regex` variable is a regular expression that matches the new email, and `old_email_expr` is a replacement expression to make the new email match the old one.

Here is another more complex example using capture groups. This example covers a situation where the email changes from `last.first@company.com` to `first.last@company.com`:
```
    {
        "new_domain": "company.com",
        "old_domains": ["company.com"],
        "email_transformation": {"new_email_regex": "([a-z]+)\.([a-z]+)", "old_email_expr": "\2.\1"},
    }
```

## Motivation and Context

A typical use-case where for this feature would be if a company were to change email domains. This feature saves the administrator time since they do not have to manually migrate users. The administrator also does not have to worry about timing a migration to match the time that the domain changes over. Artemis updates the user the first time they log in with their new email automatically.

## How Has This Been Tested?

Unit tests have been written and are passing. Feature has been tested in a live nonprod environment.

## Types of changes

## Checklist


backend/lambdas/api/authorizer/authorizer/handlers.py
```diff
@@ -29,6 +29,8 @@
 # By default give new users access to scan the demo repo
 DEFAULT_SCOPE = json.loads(os.environ.get("ARTEMIS_DEFAULT_SCOPE", "[]"))
 
+EMAIL_DOMAIN_ALIASES = os.environ.get("EMAIL_DOMAIN_ALIASES", "[]")
```
pizen, line 32: I think the `json.loads()` should happen here (like it does with `DEFAULT_SCOPE` above) rather than down in the `_get_update_or_create_user` method.

backend/lambdas/api/authorizer/authorizer/handlers.py
```diff
@@ -165,25 +170,83 @@ def _verify_claims(claims: dict):
...
+                    user = _get_user(old_email)
+
+                    # If a user is found with an old email (and was not soft-deleted), update the email and return user
+                    if user and not user.deleted:
+                        user.email = email
+                        user.save()
+                        return user
+
+    # Create the user since no match has been found at this point
+    return _create_user(email)
```
pizen, line 217: I think this might need to go inside a database transaction. The old use of `User.objects.get_or_create` was safe but I think this now isn't without a transaction.

backend/lambdas/api/authorizer/tests/test_get_user.py
```diff
@@ -0,0 +1,148 @@
...
+        "old_domains": ["company1.com"],
+        "email_transformation": {"new_email_regex": "[.]", "old_email_expr": "_"},
+    },
+    {
+        "new_domain": "company2.com",
+        "old_domains": ["company2.com"],
+        "email_transformation": {"new_email_regex": "([a-z]+)\.([a-z]+)", "old_email_expr": "\\2.\\1"},
+    },
+    {"new_domain": "newcompany.com", "old_domains": ["company.com"]},
+]
```
pizen, line 20: Would it make sense to add a test case that both changes the domain name and transforms the email?

pizen (changes requested):

backend/lambdas/api/authorizer/authorizer/handlers.py
pizen: https://docs.djangoproject.com/en/3.2/topics/db/transactions/

backend/lambdas/api/authorizer/authorizer/handlers.py
jlegarreta: Fair enough, done

backend/lambdas/api/authorizer/authorizer/handlers.py
```diff
@@ -165,25 +172,85 @@ def _verify_claims(claims: dict):
...
+def _update_login_timestamp(user: User) -> None:
+    """
+    Set the last login timestamp
+    """
     user.last_login = datetime.utcnow().replace(tzinfo=timezone.utc)
     user.save()
 
-    # Return the user
+
+@transaction.atomic
```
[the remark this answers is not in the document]
jlegarreta: I thought I would just put the decorator there in case in the future this method was used outside of the other method. I suppose if such a time comes, we can always just add it back.

backend/lambdas/api/authorizer/authorizer/handlers.py
```diff
@@ -165,25 +172,85 @@ def _verify_claims(claims: dict):
...
+                    if user and not user.deleted:
+                        user.email = email
+                        user.save()
+                        return user
+
+    # Create the user since no match has been found at this point
+    return _create_user(email)
+
+
+@transaction.atomic
```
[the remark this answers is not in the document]
jlegarreta: I suppose not, I was just covering my bases by putting write operations in transactions, but this particular one is only a single query.

backend/lambdas/api/authorizer/tests/test_get_user.py
jlegarreta: I went ahead and added a new test case

pizen (approved):

PR #55, opened 2023-06-02
Improve GitHub repo pagination in orchestrator

breedenc:
## Description

This PR improves the way the Artemis orchestrator enumerates all accessible repos in a GitHub org via GitHub's API.  

- Reduces the occurrence of API timeouts by decreasing the requested page size from 100 to 75 items
- Implements error handling for API timeouts - now, if one does occur, the request will be re-queued at the same cursor
- Updates error handling for rate limiting errors from GitHub
  - In some cases, GitHub's API responses that indicate an exceeded rate limit come with a status code of 200. Logic added to detect these errors in spite of the non-error status code.

## Motivation and Context
The previous behavior resulted in frequent timeouts from GitHub's GraphQL API. These timeouts were not handled in the orchestrator's code, so they would ultimately result in an incomplete list of repos being queued for scanning.

## How Has This Been Tested?
These changes have been tested in a live nonprod environment to ensure that the number of repos queued for scanning is consistently the expected number for a given GitHub org. Additionally, unit tests for GitHub response-parsing logic have been added/updated as appropriate.

## Types of changes

## Checklist


orchestrator/lambdas/layers/heimdall_repos/heimdall_repos/github_utils.py
```diff
@@ -60,12 +61,16 @@ def _query_github_api(self, query: str) -> str:
...
             timeout=DEFAULT_API_TIMEOUT,
         )
-        if response.status_code != 200:
-            error_response = self._analyze_error_response(response)
+
+        response_text = self._get_response_text(response)
+
+        # cannot depend on status code alone because rate limit errors have a 200 code
+        if response.status_code != 200 or self._check_for_errors_in_response(response_text):
+            error_response = self._analyze_error_response(response, response_text)
```
pizen, line 69: Could we use the `/rate_limit` API endpoint to see definitively if the limit has been exceeded? https://docs.github.com/en/rest/rate-limit?apiVersion=2022-11-28#get-rate-limit-status-for-the-authenticated-user. Or, alternatively, check this API before making a request to ensure that there is capacity within the limit to be successful?

pizen: Does the GitHub GraphQL API return the `X-RateLimit-Remaining`, `X-RateLimit-Reset`, and `Retry-After` response headers to indicate status related to the rate limit?

breedenc: Yes - all of these would work for checking rate limit status specifically. I'm thinking it would be an improvement to add a call to `/rate_limit` to `_is_message_rate_abuse()` (or create a separate function to check for this e.g. `_check_rate_limit_status()`).

However, I think there is still value in this generic implementation of `_check_for_errors_in_response()`, because we'd want it to catch any sort of error that GitHub's API sends back with a non-error status code (rate limiting is just the one I noticed when testing, but that's not to say it's the only one).

breedenc: Also, I can update the comment to something more general like-
`# cannot depend on status code alone because some errors have a 200 code (e.g. rate limit)`

breedenc: @pizen implemented a check of `X-RateLimit-Remaining` in `_analyze_error_response()` while retaining the generic check as well.

pizen (approved):

PR #62, opened 2023-07-31
Update veracode dependencies

jlegarreta:
## Description

Updated the following components:
Debian Bullseye (11) -> Bookworm (12)
Ant 1.10.12 -> 1.10.13
Golang 1.16.3 -> 1.20.6
Gradle 7.0 -> 8.2.1
Maven 3.8.8 -> 3.9.3
Node 14.16.1 -> 18.17.0
OpenJDK 11.x -> 17.0.8
PHP 7.4.22 -> 8.2.8
Ruby 2.7.2 -> 3.1

govendor package removed. Depreciated in favor of go modules since 2019.

Pinned docker images to hashes, pinned apt packages to versions.

## Motivation and Context

Several components have not been updated in a long time. This also fixes a bug that was present in the old version NPM that was causing certain scans to fail (see https://github.com/npm/npm/issues/15376).

## How Has This Been Tested?

Tested locally and ran a scan in a development environment.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] My code follows conforms to the coding standards.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #65, opened 2023-10-26
Update Veracode dependencies and python3 base image

breedenc:
## Description

- Veracode Dockerfile:
  - Updated Veracode dependencies
  - For packages sourced from package managers, pinned them to minor versions, rather than specific patches
  - Pinned base images to `major.minor` versions, rather than specific SHAs, to accept patch-level updates
  - Sources maven from a `https://repo.maven.apache.org`, which keeps previous versions available (unlike `https://downloads.apache.org`)

- python3 Dockerfile:
  - Pinned base image to minor version `python3.11`, rather than major version `python3`

### Veracode image

The Veracode container image was failing to build because some of its highly specific package versions were no longer available to `apt-get`, and because its specified version of Maven was no longer available at `https://downloads.apache.org`. This PR undoes some of the work from #62 in the interest of automatically accepting patch-level updates, reducing the frequency of required image maintenance.

### python3 image

When the `python3` image shifted to Python 3.12, the image would no longer build due to: https://github.com/aio-libs/aiohttp/issues/7739

## How Has This Been Tested?

- Images now build successfully locally and in CI
- Change has been tested in a live nonprod environment to ensure continued scan functionality for plugins whose runtime environment was updated

## Types of changes

## Checklist

### DIFF
--- a/backend/Dockerfiles/Dockerfile.python3
+++ b/backend/Dockerfiles/Dockerfile.python3
@@ -1,4 +1,4 @@
-FROM python:3-alpine
+FROM python:3.11-alpine
 
 ARG MAINTAINER
 LABEL maintainer=$MAINTAINER
--- a/backend/Dockerfiles/Dockerfile.python3
+++ b/backend/Dockerfiles/Dockerfile.python3
@@ -1,3 +1,5 @@
+# Pinning to Python 3.11 because checkov requires aiohttp, and aiohttp is not yet functional with Python 3.12
+# This issue is tracked here: https://github.com/aio-libs/aiohttp/issues/7739
 FROM python:3.11-alpine
 
 ARG MAINTAINER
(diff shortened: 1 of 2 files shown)


backend/Dockerfiles/Dockerfile.python3
```diff
@@ -1,4 +1,4 @@
-FROM python:3-alpine
+FROM python:3.11-alpine
```
g-marconet, line 1: Is there some mechanism we have to come back to this?

If we don't have something outside the repo, maybe we can comment the reason that's in the description. Someone in the future might not look at this PR and it won't be immediately apparent why it's pinned to `3.11`

breedenc: Fair, I'll add a comment

g-marconet (dismissed):
LGTM

g-marconet (approved):

PR #69, opened 2023-11-06
Bugfix: Update localstack init for localstack v2

breedenc:
## Description

Updates `docker-compose.local.yml` to use localstack's [currently-supported initialization hooks](https://docs.localstack.cloud/references/init-hooks/), enabling automatic creation of localstack infrastructure for local development.

## Motivation and Context

The `docker-entrypoint-initaws.d` initialization location was [removed](https://github.com/localstack/localstack/issues/8127#issuecomment-1505615761) in localstack v2, so the localstack infrastructure specified by `initialize.sh` was not being created automatically upon initialization of the localstack container.

## How Has This Been Tested?

Local testing (this file is only used when running Artemis locally).

## Types of changes

## Checklist

### DIFF
--- a/backend/localstack/initialize.sh
+++ b/backend/localstack/initialize.sh
@@ -1,3 +1,5 @@
+#!/bin/bash
+
 awslocal sqs create-queue --queue-name artemis-task-queue
 awslocal sqs create-queue --queue-name artemis-priority-task-queue
 awslocal sqs create-queue --queue-name artemis-callback-queue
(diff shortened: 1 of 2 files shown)


backend/localstack/initialize.sh
```diff
@@ -1,3 +1,5 @@
+#!/bin/bash
```
breedenc, line 1: Script was failing without the shebang. See https://stackoverflow.com/a/27608363.

g-marconet (approved):
LGTM

PR #77, opened 2023-11-15
Have SBOM errors/alerts/debug bubble up to scan result

g-marconet:
With this change, SBOM errors/alerts/debugs will bubble up to the scan result.

## Description
- SBOM errors/alerts/debugs were going unreported as SBOM results are delivered via S3 rather than through the database like other plugin results. This PR allows SBOM errors/alerts/debugs and SBOM status to be returned in the normal scan result.
- Adds an 8hr timeout to `veracode_sbom`

## Motivation and Context
- We wanted to add a timeout to `veracode_sbom`, as we had run into issues with some scans looping infinitely.
- With a timeout and prior to this PR, we would get no feedback about the SBOM scan timing out.
- Now, a notification about the timeout will be visible in the API results and in the Artemis web UI.

## How Has This Been Tested?
- Tests pass
- Working locally
- Working in nonprod

## Types of changes

## Checklist


backend/lambdas/generators/json_report/json_report/results/sbom.py
```diff
@@ -0,0 +1,22 @@
+from artemisdb.artemisdb.consts import PluginType
+from artemisdb.artemisdb.models import Scan
+from json_report.results.results import PLUGIN_RESULTS, PluginErrors
+
+
+def get_sbom(scan: Scan) -> PLUGIN_RESULTS:
+    """
+    Unify the output of sbom plugins. We deliberately omit the `details` property in the database
+    (since SBOM results are stored in s3), so this returns no scan results or summary
```
breedenc, line 9: This comment indicates why we return no `results` or `summary` but does not mention why `success` is hard-coded to `True`. I think I understand why, but would it be good to add this one since the others are explained?

backend/lambdas/generators/json_report/json_report/results/sbom.py
```diff
@@ -0,0 +1,22 @@
...
+    NOTE: unit tests are located at api/tests/test_generate_report.py
+    Inspect tests for expected output format.
+    :param scan: django object of Artemis repo scan
+    :return: dictionary of None for findings, list of errors, boolean success, and None for summary
+    """
+
+    errors = PluginErrors()
+
+    plugin = _empty = object()
```
breedenc, line 18: Can this be omitted since this generator does not have the `if plugin is _empty` block that the other generators have?

backend/lambdas/generators/json_report/tests/test_generate_report.py
```diff
@@ -246,6 +247,18 @@
     end_time=datetime(year=2020, month=2, day=19, hour=15, minute=1, second=55, tzinfo=timezone.utc),
 )
 
+TEST_VERACODE_SBOM = PluginResult(
+    plugin_name="Veracode SBOM",
+    plugin_type="sbom",
+    success=True,
+    details=[],
+    errors=[],
```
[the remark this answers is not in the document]
g-marconet: Agreed

backend/lambdas/generators/json_report/json_report/results/sbom.py
```diff
@@ -0,0 +1,22 @@
+from artemisdb.artemisdb.consts import PluginType
+from artemisdb.artemisdb.models import Scan
+from json_report.results.results import PLUGIN_RESULTS, PluginErrors
+
+
+def get_sbom(scan: Scan) -> PLUGIN_RESULTS:
+    """
+    Unify the output of sbom plugins. We deliberately omit the `details` property in the database
+    (since SBOM results are stored in s3), so this returns no scan results or summary
+    NOTE: unit tests are located at api/tests/test_generate_report.py
```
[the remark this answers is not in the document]
g-marconet: Good catch, will fix in other report generators

backend/lambdas/generators/json_report/json_report/results/sbom.py
g-marconet: Will add it.

Basically, there is no SBOM scan state that would cause us to want `success` to be `False`. The SBOM scan results are not stored here (and are just informative anyway. I don't think one can "fail" an SBOM scan) and it would not make sense to fail a scan when an SBOM scan errors out.

backend/lambdas/generators/json_report/json_report/results/sbom.py
g-marconet: Good catch

backend/engine/processor/processor.py
```diff
@@ -145,14 +145,19 @@ def process_plugins(self, images: dict, services: dict) -> None:
...
                     logger.info("Plugin %s completed, updating results", plugin)
-                    if results.type != "sbom":
-                        if results.type == "vulnerability":
-                            process_vulns(results, self.scan.get_scan_object(), plugin)
-                        self.scan.create_plugin_result_set(start_time, results)
-                        self._cache_results(results)
-                    else:
-                        # Process SBOM results
+
+                    if results.type == "sbom":
```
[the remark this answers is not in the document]
g-marconet: Agreed. Wanted to minimize changes, but will gladly make the change if others want it too.

breedenc (approved):

PR #79, opened 2023-12-08
SBOM storage formatting fix

breedenc:
## Description

This change ensures that license IDs in SBOM reports are reported with the key `id`, rather than `license_id`. The `id` key was in the original spec (and is the format that Artemis' UI expects); however, this was unintentionally changed to `license_id` by #48.

## Motivation and Context

#48 changed the way Artemis stores SBOM-generated dependency trees on the backend (from database to S3 bucket storage). It also unintentionally introduced a change to the format of SBOM results.

When SBOM reports were generated on-the-fly from the database, the textual representation of license objects was generated by [this to_dict() function](https://github.com/WarnerMedia/artemis/blob/670f823251cb2d77436db3ef52e77f0a3b56637f/backend/libs/artemisdb/artemisdb/artemisdb/models.py#L958-L959), which represented the `license_id` field with the `id` key.

However, when reports were written as JSON files directly to an S3 bucket, this `to_dict()` representation was not used, and this field was instead stored with the `license_id` key. A field name change constitutes a breaking change, and created validation issues when downloading SBOM reports via the UI.

This change also introduces a unit test which validates that the code which parses Veracode SBOM output returns JSON in the expected schema for submission to the S3 bucket.

## How Has This Been Tested?

 - This change has been tested in a live non-production environment
 - This change has been tested with a new unit test

## Types of changes

## Checklist

### DIFF
--- a/backend/engine/plugins/veracode_sbom/main.py
+++ b/backend/engine/plugins/veracode_sbom/main.py
@@ -160,7 +160,7 @@ def lookup_licenses(coord1: str, coord2: str, version: str, libs: list) -> list:
                     continue
 
                 for license in v["licenses"]:
-                    licenses.append({"license_id": license["name"], "name": license["license"]})
+                    licenses.append({"id": license["name"], "name": license["license"]})
 
     return licenses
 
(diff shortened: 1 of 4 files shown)


g-marconet (approved):
LGTM

PR #92, opened 2024-02-14
Update Owasp Dependency Check Plugin

Kamsiy:
## Description

**What Changed?**

__This PR:__
- Updates the docker command that builds java projects
- Updates the plugin to only run when a pom.xml file is detected in the working directory
- Adds functionality to parse errors reported in the scanner's JSON output
- Disables some of the Node and Ruby scanners that run with `dependency-check`
- Adds in unit tests for the owasp plugin

## Motivation and Context

This PR fixes the build and reporting issues in the `dependency-check` plugin. Currently, the scanner is unable to build pom.xml files or generate an output due to different reasons. For example:

1. **Docker Command Issue:** This command fails to run because, there is no container image named [`plugin` in this project](https://github.com/WarnerMedia/artemis/blob/main/backend/engine/plugins/owasp_dependency_check/main.py#L30). It also fails because, [the work directory used in the command is empty](https://github.com/WarnerMedia/artemis/blob/main/backend/engine/plugins/owasp_dependency_check/main.py#L33). Because of this, the maven build would fail because it's running on an empty folder.
2. **The vulnerability fields:** the parser for OWASP  is missing the ["inventory" section](https://github.com/WarnerMedia/artemis/blob/main/backend/engine/plugins/bundler_audit/main.py#L104-L110) in each parsed component. This was causing the vulnerability processing to fail when a report was generated by `dependency-check`

## How Has This Been Tested?

- Ran locally and in a dev environment
- Added unit tests for the owasp plugin

## Types of changes

- [ ] Documentation change

## Checklist

## Pic
Where is the bug?

### DIFF
--- a/backend/engine/plugins/owasp_dependency_check/main.py
+++ b/backend/engine/plugins/owasp_dependency_check/main.py
@@ -6,16 +6,19 @@ import json
 import os
 import subprocess
 
+from typing import Any
 from engine.plugins.lib import utils
 
 log = utils.setup_logging("owasp_dependency_check")
 
 
-def attempt_maven_build(repo_path, scan_working_dir, java_versions, engine_id):
+def attempt_maven_build(repo_path: str, scan_working_dir: str, java_versions: list, engine_id: str) -> dict:
     """
     Blindly cycle through the versions of java available.
     :param repo_path: path mounted from engine
+    :param scan_working_dir: path to the cloned repo
     :param java_versions: list of java versions to try
+    :param engine_id: containerID that stores the shared volumes
     :return: dict
     """
     for version in java_versions:
@@ -59,11 +62,11 @@ def attempt_maven_build(repo_path, scan_working_dir, java_versions, engine_id):
     return {"build_status": False, "build_debug": None, "java_version": None}
 
 
-def git_clean_repo(path):
+def git_clean_repo(path: str) -> bool:
     """
     Clean repo between java build attempts
     :param path: current working directory to run command
-    :return: string
+    :return: bool
     """
     r = subprocess.run(
         ["git", "clean", "-f", "-d", "-x"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, cwd=path, check=False
@@ -71,7 +74,7 @@ def git_clean_repo(path):
     return r.returncode == 0
 
 
-def run_owasp_dep_check(repo_path, owasp_path, repo_name):
+def run_owasp_dep_check(repo_path: str, owasp_path: str, repo_name: str) -> dict:
     """
     Run the dependency-check.sh script against each built jar
     :param repo_path: path to work directory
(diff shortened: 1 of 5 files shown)


backend/engine/plugins/owasp_dependency_check/main.py
```diff
@@ -101,15 +114,28 @@ def parse_scanner_output_json(repo_path, returncode):
...
+        results = parse_vulnerabilities(data)
+        errors = parse_errors(data)
+
+    success = success & len(results) == 0
+    return {"output": results, "errors": errors, "success": success}
+
+
+def parse_vulnerabilities(data):
+    """
+    Parse Vulnerabilities reported by owasp-dependency-check in in JSON report
```
faizan12123, line 133: quick typo fix "in in"

backend/engine/plugins/owasp_dependency_check/main.py
```diff
@@ -101,15 +114,28 @@ def parse_scanner_output_json(repo_path, returncode):
...
     with open(f"{repo_path}dependency-check-report.json") as json_file:
         data = json.load(json_file)
+        results = parse_vulnerabilities(data)
+        errors = parse_errors(data)
+
+    success = success & len(results) == 0
+    return {"output": results, "errors": errors, "success": success}
+
+
+def parse_vulnerabilities(data):
```
faizan12123, line 131: if possible, insert expected data type for parameter and expected data type to be returned to make types more strict.

backend/engine/plugins/owasp_dependency_check/main.py
```diff
@@ -119,10 +145,53 @@ def parse_scanner_output_json(repo_path, returncode):
...
+                            "advisory_ids": [
+                                vuln["name"],
+                            ],
+                        },
                     }
                 )
+    return results
+
+
+def parse_errors(data):
```
faizan12123, line 159: Same as comment above about strictly setting types (if possible)

backend/engine/plugins/owasp_dependency_check/main.py
```diff
@@ -119,10 +145,53 @@ def parse_scanner_output_json(repo_path, returncode):
...
+    owasp_errors = data["scanInfo"].get("analysisExceptions", [])
+
+    for error in owasp_errors:
+        errors.append(error["exception"]["message"])
+
+    return errors
 
-    return {"output": results, "errors": [], "success": success}
+
+def pom_exists(repo_path):
```
faizan12123, line 173: same as types comment above

faizan12123 (changes requested):
Ty for leaving well written comments on each function! Just had some quick changes/feedback.

faizan12123 (commented):
It may be worth to add a feature to allow for specifically allowing users to use the include dev dependencies feature. This can be done in this PR or future PR, but since we do that for some other plugins, might as well add support for that in this plugin.

faizan12123 (approved):
LGTM

PR #98, opened 2024-03-14
Update the Service API to support Bitbucket Cloud API v1

Kamsiy:
## Description

- Updates the services endpoint to support Bitbucket Cloud API v1
- Adds the following environment variables to the services lambda. These variables are needed to create a valid HTTP request
      ```
      ARTEMIS_REVPROXY_DOMAIN_SUBSTRING,
      ARTEMIS_REVPROXY_SECRET,
      ARTEMIS_REVPROXY_SECRET_REGION  
      ```
- Updates a broken import for `repo.utils.aws`. This import is not allowed because this module is not a shared layer

## Motivation and Context

The `/users/` endpoint is not available in the rest api for Bitbucket Cloud v1

## How Has This Been Tested?

Tested locally and in a dev environment

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/backend/lambdas/api/system_services/system_services/util/service.py
+++ b/backend/lambdas/api/system_services/system_services/util/service.py
@@ -123,27 +123,32 @@ class Service:
         if REV_PROXY_DOMAIN_SUBSTRING and REV_PROXY_DOMAIN_SUBSTRING in self._service["url"]:
             headers[REV_PROXY_SECRET_HEADER] = GetProxySecret()
             revproxy = True
+        try:
+            query = ""
+            if self._org is not None:
+                query = 'organization(login: "%s") { login }' % self._org
+            else:
+                query = "viewer { login }"
 
-        if self._org is not None:
-            try:
-                response = requests.post(
-                    url=self._service["url"],
-                    headers=headers,
-                    json={"query": 'organization(login: "%s") { login }' % self._org},
-                    timeout=3,
-                )
-                if response.status_code == 200:
-                    self._auth_successful = True
-                    self._auth_type = AuthType.SVC
-                else:
-                    self._auth_successful = False
-                    if revproxy and response.status_code == 401 and response.text == "key is invalid":
-                        # Request did not make it through the reverse proxy
-                        self._reachable = False
-                        self._error = "Reverse proxy authentication failure"
-            except requests.ConnectionError:
-                self._reachable = False
+            response = requests.post(
+                url=self._service["url"],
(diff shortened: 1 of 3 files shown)


backend/lambdas/api/system_services/system_services/util/service.py
```diff
@@ -83,14 +83,18 @@ def get_services(cls, email: str, services_dict: dict):
...
+
         if key is None:
             self._error = "Unable to retrieve key"
         if self._service["type"] == ServiceType.GITHUB:
             self._test_github(key)
         elif self._service["type"] == ServiceType.GITLAB:
             self._test_gitlab(key)
-        elif self._service["type"] == ServiceType.BITBUCKET:
-            self._test_bitbucket(key)
+        elif self._service["type"] == ServiceType.BITBUCKET and not bitbucket_v1:
```
breedenc, line 94: Would it make sense to split `BITBUCKET` into two separate service types? If the APIs vary enough that we have to differentiate in code, IMO they are two different types of services. Some shared functionality can still be abstracted to shared functions where possible.

backend/lambdas/api/system_services/system_services/util/service.py
```diff
@@ -226,10 +255,10 @@ class GetProxySecret:
 
     def __new__(cls):
         if not cls._secret:
-            from repo.util.aws import AWSConnect  # pylint: disable=import-outside-toplevel
+            from artemislib.aws import AWSConnect  # pylint: disable=import-outside-toplevel
 
             aws_connect = AWSConnect(region=REV_PROXY_SECRET_REGION)
-            cls._secret = aws_connect.get_key(REV_PROXY_SECRET)["SecretString"]
+            cls._secret = aws_connect.get_secret_raw(REV_PROXY_SECRET)
```
breedenc, line 261: should this use `get_secret()` instead of `get_secret_raw()`?

backend/lambdas/api/system_services/system_services/util/service.py
Kamsiy: I agree, they are different services. I'm okay with splitting them up within the services API. Because doing this throughout Artemis would be a huge lift. What do you think?

breedenc: Yeah, I think we should split them here for now, including defining them separately in that Enum you linked in your comment, but limit the scope to this at this time.

breedenc (approved):

PR #112, opened 2024-03-29
Eslint update

Kamsiy:
## Description

This PR:
-  Updates the default config for Eslint to support typescript and the latest version of ECMAScript
- Configures the security library to load the [right plugin](https://github.com/eslint-community/eslint-plugin-security?tab=readme-ov-file#usage) 
- Adds a version for ESLINT because v9.0.0 [will break the existing configuration](https://eslint.org/docs/latest/use/configure/configuration-files)
Screenshot 2024-03-29 at 10 17 59 AM 

and v9.0.0 will not support Node versions below v19 - https://eslint.org/blog/2023/11/whats-coming-in-eslint-9.0.0/#flat-config-now-the-default-and-has-some-changes

## How Has This Been Tested?

tested in a dev environment

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/backend/Dockerfiles/Dockerfile.node14
+++ b/backend/Dockerfiles/Dockerfile.node14
@@ -28,6 +28,7 @@ RUN apt-get update && \
     rm -rf /var/lib/apt/lists/* && \
     ln -s /usr/bin/python3 /usr/bin/python && \
     npm install eslint-plugin-security && \
+    npm install typescript@5.1.6 @typescript-eslint/parser@5.50.0 && \
     npm install -g eslint npm@7.8.0 && \
     pip3 install --upgrade setuptools && \
     pip3 install --upgrade pip boto3 && \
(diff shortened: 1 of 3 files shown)


backend/Dockerfiles/Dockerfile.node14
```diff
@@ -28,7 +28,8 @@ RUN apt-get update && \
     rm -rf /var/lib/apt/lists/* && \
     ln -s /usr/bin/python3 /usr/bin/python && \
     npm install eslint-plugin-security && \
-    npm install -g eslint npm@7.8.0 && \
+    npm install typescript@5.1.6 @typescript-eslint/parser@5.50.0 && \
+    npm install -g eslint@8.57.0 npm@7.8.0 && \
```
faizan12123, line 32: do we want to update the NPM version here while we are at it or is there a reason why we are keeping it at that version. Not sure if there would even be a benefit to updating npm here but wanted to open a potential discussion about it

faizan12123 (changes requested):

backend/Dockerfiles/Dockerfile.node14
Kamsiy: Newer versions of npm are not compatible with Node v14

In this release chart, it shows that Node v14.21.3 is released with npm v6.14.18. https://nodejs.org/en/about/previous-releases.

npm 7.8.0 is the latest version we can update to without any issues in Node 14

faizan12123: gotcha sounds good thank you!

faizan12123 (approved):
LGTM

PR #117, opened 2024-04-11
Add an event for application metadata

Kamsiy:
## Description

- Adds support for sending metadata events for downstream processing
- Updates the terraform configs to create the required AWS resources for the metadata event

## How Has This Been Tested?

tested in a dev environment

## Types of changes

- [ ] Documentation change

## Checklist


backend/terraform/modules/metadata_events/main.tf
```diff
@@ -0,0 +1,125 @@
...
+  actions = [
+    "logs:CreateLogGroup",
+    "logs:CreateLogStream",
+    "logs:PutLogEvents",
+  ]
+  iam_role_names = [
+    aws_iam_role.metadata-events-role.name
+  ]
+  name      = "${var.app}-metadata-events-write-logs"
+  resources = ["arn:aws:logs:*:*:*"]
```
breedenc, line 102: should this be scoped a bit more narrowly? possibly something like `"arn:aws:logs:${var.aws_region}:${data.aws_caller_identity.current.account_id}:log-group:/aws/lambda/${var.app}*:*"`

backend/libs/artemislib/setup.py
```diff
@@ -27,7 +27,7 @@
         # separately because it contains platform-dependent pre-compiled code
         "pyjwt[crypto]",
         "requests",
-        "urllib3<2",  # https://github.com/boto/botocore/issues/2926
+        "urllib3",
```
breedenc, line 30: It looks like this change was made in `backend/libs/artemislib/setup.py` rather than `backend/lambdas/events/metadata_handler/setup.py` - the `setup.py` for `metadata_handler` still seems to have `urllib3<2`.

We might not _want_ to mess with `artemislib` in an otherwise-unrelated PR. I was just suggesting testing this change in the new component.

PR #125, opened 2024-04-24
Enable specifying the allowed metadata schemes

ZoogieZork:
This enables unit tests to directly provide the allowed metadata schemes on a per-test basis, without relying on or modifying the global configuration.

## Description

`metadata.get_all_metadata()` now takes an optional loaded metadata schemes argument, which is in turn generated by the calling `metadata.load_schemes()` with the optional list of plugins to load.

## Motivation and Context

When developing custom metadata scheme plugins, we want the unit tests to limit the schemes to only our plugin so that other plugins (which might be loaded from the global configuration) won't interfere.

## How Has This Been Tested?

Modified unit tests in a custom plugin to use this facility rather than setting environment variables (which affect all tests, not just for the plugin).

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/backend/engine/metadata/metadata.py
+++ b/backend/engine/metadata/metadata.py
@@ -2,17 +2,20 @@ from typing import Tuple
 
 from metadata.util import is_name_in_patterns, load_schemes
 
-schemes = load_schemes()
+default_schemes = load_schemes()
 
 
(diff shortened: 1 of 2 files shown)


g-marconet (approved):
LGTM

PR #136, opened 2024-05-07
Split format checks into separate "style-test"

ZoogieZork:
Significantly speed up unit tests by splitting out the Black style checks into a separate test.

## Description

This significantly speeds up the backend unit tests:

```text
CI Before:  5m 35s
CI After:   2m 9s
```

* In local development, we assume that most folks will have the editor-of-choice run the formatter (i.e. Black) on save already.
* In CI where we can't rely on the pytest cache, running Black on all files at once is much faster than pytest-black since it can parallelize the checks -- pytest-black starts a new child process of Black for each individual file, which is very slow.

The `make test` target, being the "test everything" target, now includes running `style-test`.

## Motivation and Context

Running the backend unit tests was particularly slow due to the way pytest-black works -- it spawns a new Black subprocess for every file, so Black can't take advantage of parallelization.

This only updates the backend tests for now, since they are the slowest to run.

## How Has This Been Tested?

Ran tests locally and in CI.

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -326,16 +326,16 @@ end2end-test: test-deps
 	$(call run-pytest,end to end tests,-m "end2end")
 .PHONY: end2end-test
 
-test: test-deps format-test
+test: test-deps style-test
 	$(call run-pytest,all tests)
 .PHONY: test
 
-format-test:
-	@echo "${INFO}Running format checks"
+style-test:
+	@echo "${INFO}Running code style checks"
 	${PYTHON} -m pipenv install --dev
 	${PYTHON} -m pipenv run black --check .
 	@echo "${OK}"
-.PHONY: format-test
+.PHONY: style-test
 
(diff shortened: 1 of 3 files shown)


backend/Makefile
```diff
@@ -326,10 +326,17 @@ end2end-test: test-deps
...
 
-test: test-deps
+test: test-deps style-test
 	$(call run-pytest,all tests)
 .PHONY: test
 
+style-test:
+	@echo "${INFO}Running code style checks"
+	${PYTHON} -m pipenv install --dev
+	${PYTHON} -m pipenv run black --check .
```
ZoogieZork, line 336: Note: We don't need to add `--line-length 120` in this case since it's configured in `pyproject.toml`.

mdfleury-wbd (approved):
Looks good to me.

PR #146, opened 2024-05-20
Update Python dependencies (2024-05-20)

ZoogieZork:
This is the first of what will hopefully be a more regular cadence of dependency updates.

> [!NOTE]
> This only addresses the Python dependencies.  JS dependencies in the UI will addressed separately.

## Description

* Updated the Pipfile to specify more conservative version ranges based on the currently-locked versions -- this is to allow us to more carefully upgrade dependencies to the newest major versions a piece at a time.
* Documented (via comments in the Pipfile) why certain packages are locked at specific versions.
* Sorted dependency lists alphabetically.
* Subcomponents with their own `setup.py` have been updated to match the version specifiers in the Pipfile.  This is so that the version that pip bundles (at build-time) closer matches what we test and develop with. This doesn't _guarantee_ that pip will resolve to the same version as the lockfile, but it's a stopgap until we implement a better solution.
* The `urllib3<2` restrictions have been removed; `boto3` has been upgraded, so this is no longer necessary.

Dependencies removed:

* `pytest-cache`, `pytest-logging`: Now built-in to pytest.
* `pytest-black`: No longer used (style checks are now done separately since it's much, much faster that way).

## Motivation and Context

In the past, dependency updates have been haphazard and sometimes included surprise major version updates.  We want to establish a more regular cadence for planning and executing dependency updates.

## How Has This Been Tested?

Tested in nonprod environment and unit tests completed in a clean local environment.

## Types of changes

- [ ] Documentation change

## Checklist


g-marconet (approved):
LGTM

PR #166, opened 2024-06-04
Switch from python-jose to joserfc.

ZoogieZork:
## Description

* Switched JWT parsing and loading from python-jose to joserfc.
* Refactored the loading of the Cognito public keys into a separate function for easier unit testing.
* Added sanity check that at least one public key was loaded.

## Motivation and Context

python-jose is effectively unmaintained and is the source of a number of security reports in our dependencies.  Moving to joserfc will allow us to remove python-jose as a dependency altogether (in a future PR) as well as simplifies our code.

## How Has This Been Tested?

Tested in nonprod environment.

## Types of changes

- [ ] Documentation change

## Checklist


backend/lambdas/api/authorizer/tests/test_cognito.py
```diff
@@ -0,0 +1,53 @@
+import unittest
+
+from joserfc import jwk
+import responses
```
mdfleury-wbd, line 4: Is responses in our pipfile?

ZoogieZork: Good catch -- we're actually using responses in a bunch of other tests, but looks like it never got added to the Pipfile.

ZoogieZork: We'll explicitly add responses to the Pipfile as part of the next round of dependency updates.

mdfleury-wbd (approved):

PR #186, opened 2024-07-01
GHRC - Support branch rules

g-marconet:
Creates branch rules equivalents of the branch protection rule rules

**NOTE**: This is going into a ghrc 1.0.0 feature branch, not `main` 

## Description
- Creates `branch_rule_commit_signing`
- Creates `branch_rule_pull_requests`
- Creates `branch_rule_status_checks`
- Creates `branch_ruleset_bypass_actors` (equivalent to `branch_protection_enforce_admins`, but much more customizable due to the different bypass model Github uses for branch rules)
- Renames `branch_commit_signing` -> `branch_protection_commit_signing`
- Renames `branch_enforce_admins` -> `branch_protection_enforce_admins`
- Renames `branch_pull_requests` -> `branch_protection_pull_requests`
- Renames `branch_status_checks` -> `branch_protection_status_checks`
- Because of the above renames, this is going into a GHRC 1.0.0 feature branch rather than directly into main
  - This will also eventually include logic that will allow joining both rules together so that if either rule passes, the new rule will pass. For example, if either `branch_rule_commit_signing` or `branch_protection_commit_signing` would pass for a branch, this new rule would pass

## Motivation and Context
- Github has introduced branch rules as an alternative way to enforce these same controls on branches, so we should support them as well.

## How Has This Been Tested?
- Tests pass
- Working in local dev environment

## Types of changes
- [x] Documentation change

## Checklist

### DIFF
--- a/backend/engine/plugins/github_repo_health/cli/src/rules/branch_protection_commit_signing.py
+++ b/backend/engine/plugins/github_repo_health/cli/src/rules/branch_protection_commit_signing.py
@@ -2,11 +2,13 @@ from .helpers import add_metadata, severity_schema
 
 from github import GithubException
 
+NOT_PROTECTED_ERROR_MESSAGE = "Branch not protected"
 
-class BranchEnforceAdmins:
-    identifier = "branch_enforce_admins"
-    name = "Branch - Enforce Protection for Admins"
-    description = 'Branch protection rule, "Do not allow bypassing the above settings" is enabled. This enforces branch protection for admins'
+
+class BranchProtectionCommitSigning:
+    identifier = "branch_protection_commit_signing"
+    name = "Branch Protection - Require Commit Signing"
+    description = "Requires that a branch protection rule is enabled to enforce code signing"
 
     config_schema = {
         "type": "object",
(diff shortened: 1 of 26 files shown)


backend/engine/plugins/github_repo_health/cli/src/rules/branch_protection_commit_signing.py
```diff
@@ -26,7 +26,7 @@ def check(github, owner, repo, branch, config={}):
         try:
             protection_config = github.get_branch_protection(owner, repo, branch)
         except GithubException as e:
-            return add_metadata(False, BranchCommitSigning, config, error_message=e.data.get("message"))
+            return add_metadata(False, BranchProtectionCommitSigning, config, error_message=error_message)
```
breedenc, line 29: Does `error_message` get defined someplace? My IDE claims it's not, and I noticed the other rules still seem to use `e.data.get("message")` here

backend/engine/plugins/github_repo_health/cli/src/utilities/config.py
```diff
@@ -9,12 +9,13 @@
     "name": "default",
     "version": "0.0.1",
     "rules": [
```
breedenc, line 11: Is the addition of branch rulesets to the default config planned for a later update to the feature branch (once rules can be joined together)?

backend/engine/plugins/github_repo_health/cli/src/rules/branch_protection_commit_signing.py
g-marconet: Ah good catch, had added a bit to handle the case where PyGithub raises a `Branch not protected` error to fail without error, but removed it when I realized it didn't really need to be handled specially and missed this.

backend/engine/plugins/github_repo_health/cli/src/utilities/config.py
g-marconet: Yup! Once we add composite rules, we'll add composite rules that mimic the functionality of the original rules, but will succeed for either branch protection rules or branch rules.

For example, we'll have a composite rule, `branch_commit_signing`, that succeeds if `branch_protection_commit_signing` or `branch_rule_commit_signing` succeeds (and the same for the other branch protection rules)

breedenc (approved):

PR #201, opened 2024-07-17
Send Failed Messages to a DLQ (1)

Kamsiy:
## Description

Send failed messages from the `repo-queue Lambda` to the `org-dead-letter queue`

## How Has This Been Tested?

tested in a dev environment

## Types of changes

- [ ] Documentation change

## Checklist

## Pic
Embed something funny here

### DIFF
--- a/orchestrator/lambdas/repo_queue/repo_queue/repo_queue.py
+++ b/orchestrator/lambdas/repo_queue/repo_queue/repo_queue.py
@@ -1,18 +1,24 @@
 # pylint: disable=no-name-in-module, no-member
 import json
 from itertools import zip_longest
+from requests import HTTPError
 from typing import Optional, Any
 
 from aws_lambda_powertools import Logger
 from aws_lambda_powertools.utilities.typing import LambdaContext
 from botocore.exceptions import ClientError
 
-from heimdall_utils.aws_utils import get_analyzer_api_key, get_heimdall_secret, get_sqs_connection
+from heimdall_utils.aws_utils import (
+    get_analyzer_api_key,
+    get_heimdall_secret,
+    get_sqs_connection,
+    queue_service_and_org,
+)
 from heimdall_utils.env import API_KEY_LOC, APPLICATION
 from heimdall_utils.get_services import get_services_dict
 from heimdall_utils.utils import ServiceInfo, ScanOptions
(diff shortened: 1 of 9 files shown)


orchestrator/lambdas/repo_queue/repo_queue/repo_queue.py
```diff
@@ -27,19 +28,26 @@ def run(event: dict[str, Any] = None, context: LambdaContext = None, services_fi
...
+                default_branch_only,
+                plugins,
+                full_services_dict["external_orgs"],
+                data.get("batch_id"),
+                artemis_api_key,
+                data.get("redundant_scan_query"),
+                data.get("repo"),
+            )
+        except HTTPError:
+            log.warning("Unable to Process this organization. Sending task to dead-letter queue")
```
ZoogieZork, line 48: We should include some context with this log message (at least service, org, repo, error message) so that we can debug issues in the future.

ZoogieZork (changes requested):

orchestrator/lambdas/repo_queue/repo_queue/repo_queue.py
Kamsiy: Those keys are already added to the logger in a previous step: 
https://github.com/WarnerMedia/artemis/blob/main/orchestrator/lambdas/repo_queue/repo_queue/repo_queue.py#L74

Any logs statements that come after that step will have the additional keys

ZoogieZork: Oh nice, didn't see that.

I don't see `repo` in there -- is that a useful field to include?

Kamsiy: The `repo` key is not added in the top-level logger because the `repo` field changes a lot and could give false logs. For example a log could have a `repo` field but its not linked to the named repo

But it is included in the child loggers. here's a spot where the repo field is added in: https://github.com/WarnerMedia/artemis/blob/8009f8b85503939c4aaa95001cdcc8134333575c/orchestrator/lambdas/layers/heimdall_repos/heimdall_repos/github_utils.py#L276-L278

ZoogieZork (approved):

PR #210, opened 2024-07-24
Add plugin for Trufflehog v3

g-marconet:
Adds a Trufflehog v3 secrets plugin

## Description
- Adds a Trufflehog v3 secrets plugin
- Splitting off from Trufflehog v2 since the v2 custom secrets patterns are not supported in the same fashion in v3 (and there are similar analogs for all of our rules within v3)
- This PR does not surface verification status of findings. That will come later
- Changes the plain english name of the old Trufflehog plugin to `Trufflehog (Legacy)`

## Motivation and Context
- Trufflehog v3 is the latest version, but is meant to use its detectors rather than regular expressions. While it supports custom regular expressions, this support is described as "alpha", so we figured it'd be better to split off from v2.
  - That said, many of our custom regular expressions used for v2 have similar detectors in v3. We will look at data and see if it's safe to deprecate the v2 plugin if the false negative rate is low enough.

## How Has This Been Tested?
- API working locally
- UI working locally
- Tests pass locally

## Types of changes
- [ ] Documentation change

## Checklist

## Pic
Trufflehog

### DIFF
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -1,4 +1,4 @@
-SHELL := /bin/bash
+SpirunnerHELL := /bin/bash
 
 ###############################################################################
 # Output formatting helpers
@@ -208,6 +208,10 @@ PHP_SCANNER_VER := 1.0.0
 # Trivy
 TRIVY_COMMIT := 6ccc0a554b07b05fd049f882a1825a0e1e0aabe1
 TRIVY_VER := v0.49.1
+
+# Trufflehog Version
+TRUFFLEHOG_COMMIT := f37f2eff68c31620fd88bc37ff2b7d57ea0c700c
+TRUFFLEHOG_VER := 3.79.0
 
 # Snyk
 SNYK_VER := v1.889.0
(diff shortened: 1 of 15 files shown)


backend/engine/tests/test_plugin_truffle_hog.py
g-marconet, line 1: This is the old `test_plugin_trufflehog.py`, renamed. The only change to this file is adding underscores to the class name so that it matches the plugin name (`TestPluginTruffle**_**hog`)

backend/Makefile
```diff
@@ -1,4 +1,4 @@
-SHELL := /bin/bash
+SpirunnerHELL := /bin/bash
```
mdfleury-wbd, line 1: What's this?

mdfleury-wbd (approved):
LGTM

backend/Makefile
ZoogieZork: Looks like an accidental paste typo?

ZoogieZork (changes requested):
Just need a fix for the change in `backend/Makefile`.

backend/Makefile
g-marconet: Embarrassing. Accidental vim `.` that got in without me noticing while fixing the apirunner target

ZoogieZork (approved):

PR #227, opened 2024-08-01
artemislib: Fix type hints and handle unspecified bucket names

ZoogieZork:
## Description

Fixes a number of type hints in artemislib and make them more consistent.  Along with these fixes:

* Singleton instance fields are filled-in and typed.
* Cases where values can be `None` are accounted for:
  - Missing bucket names now raise `ValueError` instead of crashing unexpectedly down the line.
  - Handle cases where `ClientError`s are missing a code or message.
  - `Optional[T]` is used instead of `Union[T, None]` for consistency.
* Improved documentation.

Unit tests added and updated as necessary to cover the new cases.

> [!NOTE]
> The only cases where functionality changes are the missing bucket cases where we now raise `ValueError` early.  Almost none of the existing code that calls these functions handled the possibility of these functions failing or returning `None`, so this is an improvement by failing early at the site of the error rather than the failure happening later.  Code which _does_ handle the failure cases catch the exceptions, so nothing changes in those cases.

## Motivation and Context

This is a work-in-progress to gauge the impact of enabling [Pyright](https://microsoft.github.io/pyright/) type checks:

* What type errors are detected?
* What existing bugs / unhandled cases are revealed by type checking?
* What would it take to resolve these issues?

## How Has This Been Tested?

Testing in progress.

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/backend/Makefile
+++ b/backend/Makefile
@@ -417,7 +417,7 @@ dist/lambdas/layers/artemislib.zip: ${SHARED_LIB_SRC}
 	@echo "${INFO}Building $@"
 	mkdir -p ${LAMBDA_LAYERS_BUILD_DIR}/artemislib/python
 	${PIP} install --upgrade --target ${LAMBDA_LAYERS_BUILD_DIR}/artemislib/python --python-version ${LAMBDA_PYTHON_VER} --no-deps ${ARTEMISLIB}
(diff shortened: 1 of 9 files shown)


ZoogieZork:
Canceling this PR.  Splitting out the dependency updates into a separate PR to test independently: WarnerMedia/artemis#230

PR #228, opened 2024-08-01
Update base_images regex

Kamsiy:
## Description

The current regex for the base_images plugin does not catch images that have special characters in their names or Commands that use lowercase letters.

For example: These strings are not caught by the current regex:
```sh
# Case sensitive commands
from python as builder
FROM python as builder
from python AS builder
...

# Special characters
FROM golang:${GO_VERSION}-alpine AS b  
FROM python:3.11-slim-bullseye
FROM python:3.11-slim-bullseye AS builder
FROM kamsis.public.registry/docker/libs/python:3.11 AS python
FROM --platform=linux/amd64  python:3.9-alpine

# Both
FROM kamsis.public.registry/docker/libs/python:${PYTHON_VERSION} as python
...
```
## How Has This Been Tested?

- tested locally with egrep
- tested in a dev environment
## Types of changes

- [ ] Documentation change

## Checklist

## Pic
Embed something funny here


backend/engine/plugins/base_images/main.py
```diff
@@ -32,8 +32,9 @@ def find_from_lines(path: str) -> list:
             "--no-filename",
             "--recursive",
             "--include",
+            "--ignore-case",
             "Dockerfile*",
-            "^FROM (--platform=[^\\s+] ){0,1}[^\\s]+( AS .+){0,1}$",
+            "^FROM (--platform=){0,1}[-_.:\\${} /=a-z0-9]+",
```
g-marconet, line 37: The new regex is missing `$` for matching the end of the line. I'm not confident that matters, but since we had it before I'd err towards keeping it

mdfleury-wbd (approved):

PR #234, opened 2024-08-08
Expose Secret Validity in API

g-marconet:
Exposes secret validity in API

## Description
- Adds "validity" field to `gitsecrets`, `trufflehog` and `truffle_hog` results
  - `gitsecrets` and `truffle_hog` (legacy trufflehog) do not attempt to validate findings, so their results are always unknown
  - `trufflehog` (v3) will try to validate findings and will return either Active or Unknown
  - GHAS secrets already exposed the `validity` field
- Modifies the `json_report` generator to show validity
- Also fixes invalid types in files that are touched

## Motivation and Context
- GHAS secrets and Trufflehog v3 can attempt to validate that findings are legitimate. We want to surface this information to users

## How Has This Been Tested?
- Working in dev environment
- Tests pass

## Types of changes

- [ ] Documentation change

## Checklist

## Pic
That's valid

### DIFF
--- a/backend/lambdas/generators/json_report/json_report/results/secret.py
+++ b/backend/lambdas/generators/json_report/json_report/results/secret.py
@@ -54,7 +54,14 @@ def get_secrets(scan, params):
             # Build the full secrets results
             if s["filename"] not in secrets:
                 secrets[s["filename"]] = []
-            item = {"type": s.get("type"), "line": s.get("line"), "commit": s.get("commit")}
+
+            item = {
+                "type": s.get("type"),
+                "line": s.get("line"),
+                "commit": s.get("commit"),
+                "validity": s.get("validity"),
+            }
+
             if item not in secrets[s["filename"]]:
                 secrets[s["filename"]].append(item)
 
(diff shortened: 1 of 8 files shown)


backend/lambdas/generators/json_report/json_report/results/secret.py
```diff
@@ -35,31 +50,50 @@ def get_secrets(scan, params):
         # If the scan was run with a diff set the diff_summary
         diff_summary = scan.diff_summary
 
+    filename_dict: FilenameDictType = {}
```
g-marconet, line 53: The main rationale to moving towards this dict rather than continuing to directly edit the array is that we now need to de-duplicate based on a subset of properties.

For example, the following does not work, since `validity` should not be considered when de-duplicating:
```
result = [
  {
    "type": "...",
    "line": 1,
    "commit": "...",
    "validity": "unknown",
  }
]

# The same finding from another tool that does validity checks
item = {
  "type": "...",
  "line": 1,
  "commit": "...",
  "validity": "active",
}

if item not in result:
  # Same finding would be added multiple times
  result.append(item)
```

ZoogieZork (approved):

PR #237, opened 2024-08-13
Trigger Batch Scans every 3 months

Kamsiy:
## Description

Artemis Scans are currently stored for 180 days (6 months). This impacts Heimdall's ability to kickoff scans every 90 days (3 months) for repositories that have not changed in the past 3 months.

This PR updates Heimdall to verify that the timestamps used to query the Artemis API, is a date within 90 days from today

## How Has This Been Tested?

tested in a dev environment

## Types of changes

- [ ] Documentation change

## Checklist

## Pic
Embed something funny here

### DIFF
--- a/orchestrator/lambdas/layers/heimdall_utils/heimdall_utils/utils.py
+++ b/orchestrator/lambdas/layers/heimdall_utils/heimdall_utils/utils.py
@@ -91,7 +91,7 @@ def parse_timestamp(timestamp: Optional[str] = None) -> str:
 
     Examples:
         >>> parse_timestamp()
-        {"level":"WARNING","location":"parse_timestamp","message":"Generating Default timestamp"}
+        {"level":"DEBUG","location":"parse_timestamp","message":"Generating Default timestamp"}
         "2024-04-24T22:35:36Z"  # Output will vary based on the current date and time
 
         >>> parse_timestamp("2024-06-24T22:50:00Z")
@@ -100,7 +100,6 @@ def parse_timestamp(timestamp: Optional[str] = None) -> str:
     Notes:
         - The function uses the current system time to calculate the 3-month offset.
         - All returned timestamps are in UTC (denoted by the 'Z' suffix).
-        - The function logs a warning message when generating a default timestamp.
     """
     try:
         if timestamp and is_valid_timestamp(timestamp):
@@ -108,7 +107,7 @@ def parse_timestamp(timestamp: Optional[str] = None) -> str:
     except (TypeError, ValueError):
         log.error("Timestamp is invalid")
 
-    log.warning("Generating Default timestamp")
+    log.debug("Generating Default timestamp")
     default_timestamp = get_default_datetime()
     result = default_timestamp.timestamp()
 
--- a/orchestrator/lambdas/layers/heimdall_utils/heimdall_utils/utils.py
+++ b/orchestrator/lambdas/layers/heimdall_utils/heimdall_utils/utils.py
@@ -112,10 +112,14 @@ def parse_timestamp(timestamp: Optional[str] = None) -> str:
 
 
 def is_valid_timestamp(timestamp: str) -> bool:
-    default_timestamp = get_datetime_from_past(90)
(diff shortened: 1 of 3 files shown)


orchestrator/lambdas/layers/heimdall_utils/heimdall_utils/utils.py
```diff
@@ -96,18 +100,28 @@ def parse_timestamp(timestamp: Optional[str] = None) -> str:
...
+    return current_timestamp > default_timestamp
+
+
+def get_default_datetime() -> datetime:
     current_timestamp = datetime.now()
     three_months_ago = current_timestamp - timedelta(days=90)
-    result = three_months_ago.timestamp()
-
-    return time.strftime("%Y-%m-%dT%H:%M:%SZ", time.gmtime(result))
+    return three_months_ago
```
g-marconet, line 127: There's a lot implicit about this logic. Can we clear it up a little?

Parsing through the whole file, it seems like `is_valid_timestamp` is basically just testing if the given timestamp is more recent than 90 days ago, but just from reading `is_valid_timestamp` that's not immediately apparent. 

IMO, that's due to `get_default_datetime()` being a not very descriptive function name. Could we maybe refactor it to `get_90_days_ago_datetime` or even `get_datetime_from_past(days_ago)` (and then pass it `90`)

orchestrator/lambdas/layers/heimdall_utils/heimdall_utils/utils.py
```diff
@@ -96,18 +100,28 @@ def parse_timestamp(timestamp: Optional[str] = None) -> str:
...
         - All returned timestamps are in UTC (denoted by the 'Z' suffix).
-        - The function logs a warning message when generating a default timestamp.
     """
-    format = "%Y-%m-%dT%H:%M:%SZ"
     try:
-        if timestamp and bool(datetime.strptime(timestamp, format)):
+        if timestamp and is_valid_timestamp(timestamp):
             return timestamp
     except (TypeError, ValueError):
         log.error("Timestamp is invalid")
```
g-marconet, line 108: Nit: This error message may not be accurate anymore, since (if I understand correctly) a valid timestamp can now fail for being too old.

g-marconet, line 108: What errors are these trying to catch? Would it have been from the old `bool(datetime.strptime(timestamp, format))` line?

If the error is still a consideration, would it make more sense to handle this error in `is_valid_timestamp` so it's closer to the line that would throw the error?

orchestrator/lambdas/layers/heimdall_utils/heimdall_utils/utils.py
```diff
@@ -96,18 +100,25 @@ def parse_timestamp(timestamp: Optional[str] = None) -> str:
...
+    log.debug("Generating Default timestamp")
+    default_timestamp = get_datetime_from_past(90)
+    result = default_timestamp.timestamp()
+
+    return time.strftime(TIMESTAMP_FORMAT, time.gmtime(result))
 
-    return time.strftime("%Y-%m-%dT%H:%M:%SZ", time.gmtime(result))
+
+def is_valid_timestamp(timestamp: str) -> bool:
+    default_timestamp = get_datetime_from_past(90)
```
g-marconet, line 115: Leaving this at exactly 90 days might leave room for a race condition.
- Heimdall runs weekly, and it sees a scan for a given repo that expires 1 hour from now (so it doesn't trigger a rescan)
- An hour later, downstream services purge data due to being outside the 90 day retention period
- Until the next scan a week later, we do not have data in downstream service for the given repo

To solve this, we could set this to 83 days. Let's make sure to document why though, as 83 would definitely seem like a magic number.

orchestrator/lambdas/layers/heimdall_utils/heimdall_utils/utils.py
Kamsiy: It makes sense to move it to `is_valid_timestamp`. This was catching invalid timestamp strings. 
For example is_valid_timestamp("2024") would throw a ValueError

g-marconet (approved):

PR #238, opened 2024-08-15
cleaning up types

mdfleury-wbd:
## Description
Fixing typing issues

## How Has This Been Tested?
Tested locally

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/backend/engine/processor/vulns.py
+++ b/backend/engine/processor/vulns.py
@@ -237,7 +237,7 @@ def _process_vuln_instance(
     vuln_scan, _ = VulnerabilityScanPlugin.objects.get_or_create(vuln_instance=vuln_instance, scan=scan)
     if not vuln_scan.plugins.filter(pk=plugin.pk).exists():
         vuln_scan.plugins.add(plugin)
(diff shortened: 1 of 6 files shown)


backend/engine/processor/vulns.py
```diff
@@ -222,7 +237,7 @@ def _process_vuln_instance(
     vuln_scan, _ = VulnerabilityScanPlugin.objects.get_or_create(vuln_instance=vuln_instance, scan=scan)
     if not vuln_scan.plugins.filter(pk=plugin.pk).exists():
         vuln_scan.plugins.add(plugin)
-    if not vuln_scan.components.filter(pk=component.pk).exists():
+    if component is not None and not vuln_scan.components.filter(pk=component.pk).exists():
```
Kamsiy, line 240: Could this be simplified to:

```python
if component and not vuln_scan.components.filter(pk=component.pk).exists():
```

This would also catch if its an empty component as well. What do you think?

mdfleury-wbd: That seems like a good idea. Those things should never happen, but just in case.

Kamsiy (approved):

PR #256, opened 2024-09-05
Improve handling of API errors in artemis-scan.sh

ZoogieZork:
## Description

artemis-scan.sh is our published script for integrating Artemis scans into CI workflows, such as GitHub Actions.

This improves our handling of cases where the REST API calls return an error, particularly non-JSON responses.

Before, these conditions would cause the script to exit immediately with a jq error message. Now we provide a proper diagnostic message to aid in debugging.

Now when printing the response body, if the body text cannot be pretty-printed as JSON, we fall back to printing the raw body. We also add default values for when extracting a JSON value fails.

## Motivation and Context

We noticed failures of the script in existing CI runs but the jq errors were hiding diagnostic info.

## How Has This Been Tested?

Unit tests added. Performed a manual run against a non-prod environment.

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/backend/ci-tools/shell/artemis-scan.sh
+++ b/backend/ci-tools/shell/artemis-scan.sh
@@ -165,6 +165,13 @@ log() {
   echo "$(date -u '+%Y-%m-%dT%H:%M:%S%z')" "$1"
 }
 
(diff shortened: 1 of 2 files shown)


g-marconet (approved):
LGTM

PR #257, opened 2024-09-05
Add "Finding Details" for secrets findings

g-marconet:
Adds a "Finding Details" table to the secret finding dialog to provide more information about the finding. This is particularly useful for deduplication, where we now show "Mixed" for findings for which plugins report multiple validities.

## Description
- API no longer tries to determine validity. Instead it has a `details` field that contains the `type`, `validity`, and name of the plugin that found it.
- UI will now derive the validity for findings with multiple validity types, with the following logic:
  - If all subfindings have the same `validity`, use that `validity`
  - If any subfinding is `active`, the finding is `active`
  - If all subfindings are the same `type`, prioritize `inactive` over `unknown`
    - In this case, it's more likely that all subfindings are the same secret. If that's true, the `inactive` determination was likely made with more information than the `unknown` determination, so we go with `inactive`.
  - If there are multiple `type`s, prioritize `unknown` over `inactive`
    - In this case, it's possible the `unknown` determination is from a secret type that cannot be automatically verified, so we want to draw the user's attention to that.
- Secrets Dialog now shows the full details

## Screenshots

- New secrets tab validities, showing off how `Mixed` findings are displayed
  - Screenshot 2024-09-05 at 10 12 49 AM

- Secrets Dialog for the `Active (Mixed)` finding
  - Screenshot 2024-09-05 at 10 13 07 AM

- Secrets Dialog for the first `unknown` finding, which shows the case where no details are provided (if there's a version mismatch between UI and API)
  - Screenshot 2024-09-05 at 10 13 17 AM

- Secrets Dialog for the `Inactive (Mixed)` finding, which shows the case where the subfinding `type`s all match, so we prioritize `inactive` over `unknown`.
  - Screenshot 2024-09-05 at 10 13 22 AM

- Secrets Dialog for the `inactive` finding, which shows the case where the subfinding `validities` all match, so we do not show `mixed` (even though there are multiple results).
  - Screenshot 2024-09-05 at 10 13 31 AM

- Secrets Dialog for the `Unknown (Mixed)` finding, which shows the case where the subfinding `type`s do not match, and so we prioritize `unknown` over `inactive`.
  - Screenshot 2024-09-05 at 10 13 25 AM

- Tooltip for `Mixed` chips show the breakdown of validities and the finding types that are associated with them 
  - Screenshot 2024-09-05 at 10 13 53 AM
 

## Motivation and Context

Currently, Artemis hides Validity determinations in the API, abstracting this information from the user. This change exposes all the data to the user and presents it more intuitively.

## How Has This Been Tested?

- Working locally and in dev environment
- Tests pass

## Types of changes

- [ ] Documentation change

## Checklist

## Pic
Unlimited detail


ui/src/components/SecretValidityCell.tsx
```diff
@@ -101,6 +114,45 @@ export const SecretValidityChip = (props: SecretValidityChipProps) => {
 				/>
 			);
 	}
+
+	function getMixedChip(
+		details: ReadonlyArray<SecretDetail>,
+		tooltipDisabled: boolean | undefined
```
ZoogieZork, line 120: (not a blocker) This can be simplified as `tooltipDisabled?: boolean`

ZoogieZork (approved):

PR #268, opened 2024-09-13
Updating runner for ui tests to stronger runner

faizan12123:
## Description

Updating the runner to utilize a beefier runner to speed up unit tests
## Motivation and Context

to speed up UI unit tests as they take too long to finish
## How Has This Been Tested?

Through triggering the ci with the new runners
## Types of changes

- [ ] Documentation change

## Checklist


ui/Makefile
```diff
@@ -192,7 +192,7 @@ audit: package.json package-lock.json  ## Audit package dependencies for vulnera
 .PHONY: test
 test:  ## Run tests
 	@echo "${INFO}Running tests"
-	CI=true $(NPM) test -- --env=jsdom --coverage --silent
+	CI=true $(NPM) test -- --env=jsdom --coverage --maxWorkers=100% --silent
```
faizan12123, line 195: `maxWorkers=100%` seems to speed up tests on average by 1 minute but some people in forums have complained that this can cause instability in tests, such as tests timing out or not finishing before going onto the next test, however in my testing this did not occur at all. Thoughts?

breedenc: IMHO, 1 minute is unlikely worth the risk of the complication, especially if it's possible that the test would fail silently and move on to the next test. In that case, there is the risk of a developer not noticing that tests aren't functioning properly.

faizan12123: Removed! @breedenc

breedenc (approved):

PR #269, opened 2024-09-16
Adding new GHAS locations for secrets

mdfleury-wbd:
## Description
Adding support for new GHAS secret scanning features in locations besides code

## How Has This Been Tested?
Tested in nonprod

## Types of changes

- [ ] Documentation change

## Checklist


backend/engine/plugins/ghas_secrets/main.py
```diff
@@ -59,45 +60,18 @@ def _ghas_secrets(gh: GitHubAPI, path: str) -> Tuple[list[dict], dict]:
...
-                "type": _normalize_secret_type(alert["secret_type"]),
-                "author": author,
-                "author-timestamp": author_timestamp,
-                "validity": alert["validity"],
-                "state": alert["state"],
-                "created_at": alert["created_at"],
-            }
+            item = format_secret(gh, item_id, location, alert, author, author_timestamp)
             results.append(item)
             event_info[item_id] = {"match": alert["secret"], "type": item["type"]}
```
ZoogieZork, line 70: `format_secret()` can now return an empty `dict`, so need to check for that before accessing `item["type"]` here (or use something like `item.get("item", some_default_value)`).

backend/engine/plugins/ghas_secrets/formatter.py
```diff
@@ -0,0 +1,127 @@
...
+    "pull_request_body": True,
+    "pull_request_comment": True,
+    "pull_request_review": True,
+    "pull_request_review_comment": True,
+}
+
+
+def format_secret(
+    github: GitHubAPI, item_id: str, location: dict, alert: dict, author: str, author_timestamp: str
+) -> Any:
```
ZoogieZork, line 27: Looks like this always returns a `dict`, so we can return that instead of `Any`.  Otherwise, we'll need to check the return with `isinstance()` at the callsite.

backend/engine/plugins/ghas_secrets/formatter.py
```diff
@@ -0,0 +1,127 @@
...
+            "state": alert["state"],
+            "created_at": alert["created_at"],
+            "location": location["type"],
+            "url": url,
+        }
+
+
+def _normalize_secret_type(secret_type: str) -> str:
+    # There is a large number of secret types supported by GHAS. Some of them need to be normalized
+    # to be aligned with the types returned by other Artemis plugins.
```
ZoogieZork, line 116: Can we reword this and reformat it as a docstring?

ZoogieZork (changes requested):

backend/engine/plugins/ghas_secrets/validator.py
```diff
@@ -0,0 +1,62 @@
...
+    "discussion_comment": True,
+    "pull_request_title": True,
+    "pull_request_body": True,
+    "pull_request_comment": True,
+    "pull_request_review": True,
+    "pull_request_review_comment": True,
+}
+
+
+def validate(location: dict, path: str) -> Tuple[bool, str, str]:
```
ZoogieZork, line 23: (minor) We can use the built-in `tuple` instead of `typing.Tuple` [since we're on Python 3.9](https://docs.python.org/3/library/typing.html#typing.Tuple).

ZoogieZork (approved):

PR #271, opened 2024-09-18
Normalize Trufflehog Finding Types

g-marconet:
Normalizes Trufflehog Finding Types

## Description
- Trufflehog legacy
  - Remove "other" category. Instead, returns the type from Trufflehog directly if it doesn't fall into a bin
- Trufflehog v3
  - Bins finding types into the bins from Trufflehog legacy
    - With the exception of urlauth, which does not have a corresponding Trufflehog v3 Detector 

## Motivation and Context

- With https://github.com/WarnerMedia/artemis/pull/252 merged, `type` is not factored into de-duplication so the `other` type is just hiding useful information from the user. This PR removes that from Trufflehog Legacy
- That said, binning types is still useful in general to clean up results when a finding is found by multiple plugins, so we also add binning to the new Trufflehog plugin.

## How Has This Been Tested?

- Working in dev environment
- Tests pass

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/backend/engine/plugins/truffle_hog/main.py
+++ b/backend/engine/plugins/truffle_hog/main.py
@@ -60,7 +60,7 @@ def secret_type(reason: str) -> str:
     elif reason.startswith("Slack "):
         return "slack"
     else:
-        return reason;
+        return reason
 
 
 def commit_author(scan_path: str, commit: str) -> tuple:
--- a/backend/engine/plugins/truffle_hog/main.py
+++ b/backend/engine/plugins/truffle_hog/main.py
@@ -43,24 +43,24 @@ def main(in_args=None):
 
 
 def secret_type(reason: str) -> str:
-    stype = "other"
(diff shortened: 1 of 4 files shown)


backend/engine/plugins/truffle_hog/main.py
```diff
@@ -43,24 +43,24 @@ def main(in_args=None):
...
+        return "redis"
     elif reason.startswith("HTTP ") or reason.startswith("HTTPS "):
-        stype = "urlauth"
+        return "urlauth"
     elif reason.startswith("Slack "):
-        stype = "slack"
-    return stype
+        return "slack"
+    else:
+        return reason
```
g-marconet, line 63: Effectively, this change is just returning the original reason (instead of "other") when findings do not fall into an identified bin.

The `stype` intermediate variable seemed unnecessary to me, so I removed it to simplify things a little.

ZoogieZork (approved):

PR #272, opened 2024-09-19
Add Structured logs to the Artemis Engine, Plugins & Lambdas

Kamsiy:
### Example Logs:
Engine Context
```
```

Lambda Logs
```
```

Plugin Logs
```
```

## Types of changes

- [ ] Documentation change

## Checklist

## Pic
Embed something funny here

### DIFF
--- a/backend/engine/plugins/lib/utils.py
+++ b/backend/engine/plugins/lib/utils.py
@@ -14,13 +14,53 @@ APPLICATION = os.environ.get("APPLICATION", "artemis")
 REGION = os.environ.get("REGION", "us-east-2")
 
 
+class JSONFormatter(logging.Formatter):
+    """
+    Formats a LogRecord to a JSON encoded string.
+    """
+
+    def __init__(self):
+        super().__init__()
+        self.reserved_attrs = set(logging.LogRecord("", 0, "", 0, "", (), None).__dict__.keys())
+
+    def format(self, record: logging.LogRecord) -> str:
+        message = record.__dict__.copy()
+        log_record = {
+            "timestamp": self.formatTime(record, self.datefmt),
+            "level": record.levelname,
+            "location": f"{record.funcName}:{record.lineno}",
+            "message": record.getMessage(),
+        }
+        # Exclude reserved attributes and Add additional_fields
+        for key, value in message.items():
+            if key not in self.reserved_attrs and value:
+                log_record[key] = value
+
+        # Format Exceptions
+        if record.exc_info and record.exc_text is None:
+            record.exc_text = self.formatException(record.exc_info)
+
+        if record.exc_text:
+            log_record["exc_info"] = record.exc_text
+
+        if record.stack_info:
+            log_record["stack_info"] = self.formatStack(record.stack_info)
+
+        log_record = self._prune_null_fields(log_record)
+        return json.dumps(log_record)
+
+    def _prune_null_fields(self, records: dict) -> dict:
+        """
+        Remove any key with Null values
+        """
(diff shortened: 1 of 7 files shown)


backend/libs/artemislib/artemislib/logging.py
```diff
@@ -1,29 +1,182 @@
...
+
+    @classmethod
+    def remove_fields(cls, *args):
+        record_factory = cls._get_record_factory()
+        record_factory.remove_fields(*args)
+
+    @classmethod
+    def reset_fields(cls):
+        record_factory = cls._get_record_factory()
+        record_factory.reset_fields()
```
ZoogieZork, line 132: We should implement a context manager for adding fields and making sure those fields are removed/restored at the end of the block.  The Python logging cookbook has an example where they implement a context manager to adjust logging levels within a block; we can adapt that for saving/restoring fields: https://docs.python.org/3/howto/logging-cookbook.html#using-a-context-manager-for-selective-logging

Needing to make sure calls to `add_fields()` and `remove_fields()` are matched up tends to be brittle and difficult to review.  Something like this would be much easier:

```python
from artemislib.logging import LogFields

for plugin in self.action_details.plugins:
  with LogFields(logger, plugin=plugin):
    # The "plugin" field automatically gets restored at the end of the block
    logger.info( ... )
```

That said, for now, let's not let perfect be the enemy of good and save that for a later enhancement.

backend/docker-compose.aws.yml
```diff
@@ -13,8 +13,8 @@ services:
       options:
         awslogs-region: ${REGION}
         awslogs-group: ${LOG_GROUP}
+        awslogs-multiline-pattern: "^\\{(?:[^{}]|\\{(?:[^{}]|\\{[^{}]*\\})*\\})*\\}$$"
```
ZoogieZork, line 16: This is a scary-enough regex that we'll need a comment block to explain it.

Key points to address:
* What does this regex do?
* What are our assumptions about the formatting of the logs?
* An example of what data this regex is matching.
* Any known limitations of the regex?  What log patterns might not be detected or cause issues that we would need to work around?

backend/engine/plugins/lib/utils.py
```diff
@@ -28,6 +29,14 @@ def setup_logging(name):
     return log
 
 
+def handle_exception(exc_type, exc_value, exc_traceback):
```
ZoogieZork, line 32: Do we know what the types of these parameters are?

backend/libs/artemislib/artemislib/logging.py
```diff
@@ -1,29 +1,182 @@
...
+    for line in logs:
+        if "ERROR" in line:
+            logger.error(line)
+        if "CRITICAL" in line:
+            logger.critical(line)
+        else:
+            logger.info(line)
+
+
+def handle_exception(exc_type, exc_value, exc_traceback):
```
ZoogieZork, line 173: Same as above -- do we know what the types of the parameters are?

ZoogieZork (changes requested):

backend/engine/plugins/lib/utils.py
Kamsiy: Yup, these are the classes for all 3: `<class 'type'> <class 'Exception'> <class 'traceback'>`

backend/libs/artemislib/artemislib/logging.py
Kamsiy: Created a task to implement this

ZoogieZork (approved):

PR #273, opened 2024-09-19
Gitsecrets Plugin - Add additional types

g-marconet:
Add types to Gitsecrets results, instead of returning "other" for many different finding types

## Description
- Gitsecrets will now more often return a secret type, rather than binning things into an "other" bin
- We were using [old Trufflehog regexes](https://github.com/dxa4481/truffleHogRegexes/blob/9257f45e66e31ad3fefd9d6e2b2e09dfb83043d6/truffleHogRegexes/regexes.json), so this PR adds the types from those trufflehog regexes back in (except where an existing bin makes more sense)

## Motivation and Context

With
https://github.com/WarnerMedia/artemis/pull/252 merged, type is not factored into de-duplication so the other type is just hiding useful information from the user. This PR removes that from gitsecrets

## How Has This Been Tested?

Working in dev environment and tests pass

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/backend/engine/plugins/gitsecrets/secrets_processor.py
+++ b/backend/engine/plugins/gitsecrets/secrets_processor.py
@@ -2,7 +2,9 @@
 import os
 import re
 from string import Template
+from typing import Optional
 
+from engine.plugins.gitsecrets.secret_regex import SecretRegex
 from engine.plugins.lib.utils import setup_logging
 
 log = setup_logging("gitsecrets")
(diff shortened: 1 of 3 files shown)


backend/engine/plugins/gitsecrets/secrets_processor.py
```diff
@@ -2,56 +2,150 @@
...
-    re.compile(str(r"([a|A][p|P][i|I][_]?[k|K][e|E][y|Y](.){1,64}['|\"][0-9a-zA-Z]{32,45}['|\"])")),
-    re.compile(str(r"([s|S][e|E][c|C][r|R][e|E][t|T](.){1,64}['|\"][0-9a-zA-Z]{32,45}['|\"])")),
-    re.compile(str(r"([0-9]+-[0-9A-Za-z_]{32}\.apps\.googleusercontent\.com)")),
-    re.compile(str(r"(ya29\.[0-9A-Za-z\-_]{55})")),
-    re.compile(
-        str(
-            (
-                r"((ftp|ftps|http|https|mongodb|mongodb\+srv|postgres)"
-                r"://[^/\s:@]{3,64}:[^/\s:@]{3,64}@[a-zA-Z0-9\.-]+(:[0-9]{1,5}){0,1})"
+    SecretRegex(
```
g-marconet, line 20: Pulls the types from https://github.com/dxa4481/truffleHogRegexes/blob/9257f45e66e31ad3fefd9d6e2b2e09dfb83043d6/truffleHogRegexes/regexes.json

Which I believe is where we originally got these regular expressions

backend/engine/plugins/gitsecrets/secrets_processor.py
```diff
@@ -2,56 +2,150 @@
...
+        regex=re.compile(str(r"((mongodb|mongodb\+srv)" + URL_REGEX)),
+    ),
+    SecretRegex(
+        finding_type="postgres",
+        regex=re.compile(str(r"((postgres)" + URL_REGEX)),
+    ),
+    SecretRegex(
+        finding_type="urlauth",
+        regex=re.compile(str(r"((ftp|ftps|http|https)" + URL_REGEX)),
+    ),
```
g-marconet, line 79: These three are collapsed into one rule when we run gitsecrets itself. We break them out, but it should be equivalent.

PR #274, opened 2024-09-20
Add type hints to Engine plugin.py

ZoogieZork:
## Description

Type hint updates for backend/engine/utils/plugin.py before making further changes.

* Add and fix type hints.
* Reformat docstrings previously written as comments.

No changes to the code itself, only type hints and docstrings.

## Motivation and Context

Separating out the simple type hint and docstring fixes from a (future) PR.

## How Has This Been Tested?

No code changes, only updates to type hints and docstrings.

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/backend/engine/utils/plugin.py
+++ b/backend/engine/utils/plugin.py
@@ -4,7 +4,7 @@ import subprocess
 from dataclasses import dataclass
 from datetime import datetime, timezone
 from fnmatch import fnmatch
-from typing import Optional, Tuple
+from typing import Optional
 from urllib.parse import quote_plus
 
 import boto3
@@ -144,7 +144,7 @@ def pull_image(image: str):
     return None
 
 
-def execute_docker_pull(image, log_error) -> bool:
+def execute_docker_pull(image: str, log_error: bool) -> bool:
     """
     Executes docker pull [image]
     if the return code is not 0, there was an error.
@@ -325,6 +325,7 @@ def run_plugin(
 
     try:
         plugin_output = json.loads(r.stdout)
+        # TODO: Validate plugin output is a dict.
 
         if "event_info" in plugin_output:
             # Process event info by sending them to a locked down SQS queue so that the results can be
@@ -374,7 +375,7 @@ def run_plugin(
     )
 
 
-def process_event_info(scan, results, plugin_type, plugin_name):
+def process_event_info(scan: Scan, results, plugin_type: str, plugin_name: str):
     log.info("Processing event info")
     timestamp = get_iso_timestamp()
     if plugin_type == PluginType.SECRETS.value and SECRETS_EVENTS_ENABLED:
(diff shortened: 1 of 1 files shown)


backend/engine/utils/plugin.py
```diff
@@ -475,7 +488,10 @@ def get_secret_raw_wl(scan):
 
 
 def get_secret_al(scan):
```
Kamsiy, line 490: could we add a `Scan` type here?

backend/engine/utils/plugin.py
```diff
@@ -460,7 +469,11 @@ def queue_event(repo, plugin_type, payload):
 
 
 def get_secret_raw_wl(scan):
```
Kamsiy, line 471: and here

backend/engine/utils/plugin.py
ZoogieZork: Not yet, sadly -- this function uses a Django implicit attribute (`allowlistitem_set`) that can't be type-checked.  I added a comment so we can revisit this later:
```python
# Note: scan type is unspecified until we enable typechecking Django models.
```

backend/engine/utils/plugin.py
ZoogieZork: Same issue as above (and same implicit attribute which is the problem).

Kamsiy (approved):

backend/engine/utils/plugin.py
```diff
@@ -366,7 +375,7 @@ def run_plugin(plugin, scan, scan_images, depth=None, include_dev=False, feature
     )
 
 
-def process_event_info(scan, results, plugin_type, plugin_name):
+def process_event_info(scan: Scan, results, plugin_type: str, plugin_name: str):
```
faizan12123, line 378: I see the param `results` does not have an expected type, is it because any data type can be expected here that can be expected here?

backend/engine/utils/plugin.py
```diff
@@ -240,7 +240,15 @@ def is_plugin_disabled(settings: dict) -> bool:
     return True
 
 
-def run_plugin(plugin, scan, scan_images, depth=None, include_dev=False, features=None, services=None) -> Result:
+def run_plugin(
```
faizan12123, line 243: I see some expected data types missing here as well, any reason for that?

backend/engine/utils/plugin.py
ZoogieZork: Yeah, I left some params as untyped if I couldn't determine with certainty the possible types.

backend/engine/utils/plugin.py
ZoogieZork: `scan_images` is left untyped here since I'm not 100% certain all of the places it could be generated -- yet :)

This one is a work-in-progress that hopefully can be properly defined as a dataclass.

backend/engine/utils/plugin.py
```diff
@@ -240,7 +240,15 @@ def is_plugin_disabled(settings: dict) -> bool:
...
+    plugin: str,
+    scan: Scan,
+    scan_images,
+    depth: Optional[str] = None,
+    include_dev=False,
```
faizan12123, line 248: include_dev should just be True or False, could we explicitly call this out as expecting a `bool`?

ZoogieZork: This is actually a weird case.  The type of `include_dev` can actually be `Union[str, bool]`, with an extra potential `Unknown` in there because AFAICT it's not validated or constrained to `bool` when passed-in.

Since this PR is intended to skip over anything that would require a change of code outside this one file, this param was left as-is for now.

PR #281, opened 2024-09-25
Fix bug in trivy_sbom

Kamsiy:
## Description

The package names stored for trivy_sbom results are not correct. For example, packages with a group field are stored as: `@balenadockerignore` and not `@balena/dockerignore`

#### This PR:
Adds a divider between the component group and the component name for a detected package.

Example Cyclone DX component for https://www.npmjs.com/package/@balena/dockerignore
```
{
      "bom-ref": "pkg:npm/%40balena/dockerignore@1.0.2",
      "type": "library",
      "group": "@balena",
      "name": "dockerignore",
      "version": "1.0.2",
      "purl": "pkg:npm/%40balena/dockerignore@1.0.2",
      "properties": [
        {
          "name": "aquasecurity:trivy:PkgID",
          "value": "@balena/dockerignore@1.0.2"
        },
        {
          "name": "aquasecurity:trivy:PkgType",
          "value": "yarn"
        }
      ]
    }
```

## Motivation and Context

This affects the `artemis-license-retriever` Lambda because it uses the package name to build URLs to query different package registries

## Types of changes

- [ ] Documentation change

## Checklist

## Pic
Embed something funny here

### DIFF
--- a/backend/engine/plugins/trivy_sbom/parser.py
+++ b/backend/engine/plugins/trivy_sbom/parser.py
@@ -19,6 +19,9 @@ def clean_output_application_sbom(output: list) -> list:
         if item["type"] == "operating-system":
             continue
         bom_ref = item["bom-ref"]
+        if "name" not in item:
+            logger.error("Unable to parse trivy component: %s", item)
+            continue
         if "group" in item:
             name = f'{item["group"]}/{item["name"]}'
         else:
--- a/backend/engine/plugins/trivy_sbom/parser.py
+++ b/backend/engine/plugins/trivy_sbom/parser.py
@@ -19,7 +19,10 @@ def clean_output_application_sbom(output: list) -> list:
         if item["type"] == "operating-system":
             continue
         bom_ref = item["bom-ref"]
-        name = item.get("group", "") + item["name"]
+        if "group" in item:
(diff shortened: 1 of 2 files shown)


backend/engine/plugins/trivy_sbom/parser.py
```diff
@@ -19,7 +19,10 @@ def clean_output_application_sbom(output: list) -> list:
         if item["type"] == "operating-system":
             continue
         bom_ref = item["bom-ref"]
-        name = item.get("group", "") + item["name"]
+        if "group" in item:
+            name = f'{item["group"]}/{item["name"]}'
+        else:
+            name = item["name"]
```
faizan12123, line 25: should we be `None` type safe here and just make it name = item.get("name") instead?

Kamsiy: If the name of the package is missing, that is a bigger issue. I'll add a check to handle a missing `name` field because other components depend on that field.

FWIW: The cyclone dx schema mentions that `name` is a required field - https://github.com/CycloneDX/specification/blob/1.6/schema/bom-1.6.schema.json#L830-L832 

But i will add a check incase its missing in the trivy result

faizan12123 (approved):
LGTM

PR #283, opened 2024-09-25
Trufflehog Plugin: Add detector verification allowlist

g-marconet:
Adds a detector verification allowlist

NOTE: The Trufflehog v3 plugin is still disabled by default

## Description

- Adds a detector verification allowlist. All detectors not on the allowlist will still be run, but without verification (so they will not make calls out to services)
- Maps "URI" detector to "urlauth" (this was missed in https://github.com/WarnerMedia/artemis/pull/271)
- Updates Trufflehog to `3.82.5`

## Motivation and Context

- Trufflehog detectors can in theory (and some in practice) have too-permissive regular expressions that could cause Artemis to send valid credentials for one service to another service. This could be an attack vector, so we only verify findings for detectors that have been determined to be safe.

## How Has This Been Tested?

- Working in dev
- Tests pass

## Types of changes

- [ ] Documentation change

## Checklist


backend/engine/plugins/trufflehog/type_normalization.py
```diff
@@ -31,6 +32,9 @@ def get_type_normalization_table() -> dict[str, str]:
     for name in _SLACK_TYPES:
         table[name] = "slack"
 
+    for name in _URLAUTH_TYPES:
+        table[name] = "urlauth"
+
```
g-marconet, line 37: This detector type was missed in https://github.com/WarnerMedia/artemis/pull/271

ZoogieZork (approved):

PR #290, opened 2024-09-27
Fix GitSecrets Patterns

g-marconet:
Fixes gitsecrets patterns

## Description

- Removes escape slash for forward slashes (there's a warning in the git-secrets output otherwise. Small thing)
- Adds double brackets around standalone `[:space:]`. `[:space:]` is equivalent to `\s` when it's already in a set, but needs double brackets (`[[:space:]]`) when it is not in a set.

## Motivation and Context

- https://github.com/WarnerMedia/artemis/pull/288 had a bug, this fixes that

## How Has This Been Tested?

- Working in dev
  - Bug wasn't discovered in dev before due to the `dist/docker/python3` make target not rebuilding when the secret patterns change. I've manually triggered a rebuild and it works

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/backend/Dockerfiles/data/secret-patterns
+++ b/backend/Dockerfiles/data/secret-patterns
@@ -10,7 +10,7 @@ EAACEdEose0cBA[0-9A-Za-z]{16,255}
 [s|S][e|E][c|C][r|R][e|E][t|T](.){1,64}['|"][0-9a-zA-Z]{32,45}['|"]
 [0-9]+-[0-9A-Za-z_]{32}\.apps\.googleusercontent\.com
 ya29\.[0-9A-Za-z\-_]{55}
-(ftp|ftps|http|https|mongodb|mongodb\+srv|postgres):\/\/[^/:@[:space:]]{3,64}:[^/:@[:space:]]{3,64}@[a-zA-Z0-9\.-]+(:[0-9]{1,5}){0,1}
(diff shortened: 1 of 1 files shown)


backend/Dockerfiles/data/secret-patterns
```diff
@@ -10,7 +10,7 @@ EAACEdEose0cBA[0-9A-Za-z]{16,255}
 [s|S][e|E][c|C][r|R][e|E][t|T](.){1,64}['|"][0-9a-zA-Z]{32,45}['|"]
 [0-9]+-[0-9A-Za-z_]{32}\.apps\.googleusercontent\.com
 ya29\.[0-9A-Za-z\-_]{55}
-(ftp|ftps|http|https|mongodb|mongodb\+srv|postgres):\/\/[^/:@[:space:]]{3,64}:[^/:@[:space:]]{3,64}@[a-zA-Z0-9\.-]+(:[0-9]{1,5}){0,1}
+(ftp|ftps|http|https|mongodb|mongodb\+srv|postgres)://[^/:@[:space:]]{3,64}:[^/:@[:space:]]{3,64}@[a-zA-Z0-9\.-]+(:[0-9]{1,5}){0,1}
```
g-marconet, line 13: This doesn't change any functionality, but grep warns about these lines otherwise (and we filter those warnings out, so it's just a small optimization and to clean up output)

faizan12123 (approved):
LGTM

PR #299, opened 2024-10-08
New plugin runner: boxed

ZoogieZork:
## Description

This introduces a new method of running plugins within a container, named `boxed`.

Instead of relying on the system Python and dependencies in the target container, this bundles the plugins, dependencies, and a "portable" Python distribution into a single self-contained executable.

This means that the target container no longer needs a system Python installed and we don't need to make sure the dependencies are installed in the container (which can get out of sync with the engine dependencies).

The only new requirement of the target container is that it must be a glibc-based distribution since the bundle contains a glibc-based Python and shared libraries (i.e., this rules out Alpine).  The plugin documentation has been updated to list the container requirements for both old and new runners.

The gosec and enry ("technology_discovery") plugins have been switched to this new runner.  As a result, the Dockerfile.golang has been greatly simplified: We no longer need to install Python or Python dependencies.  We did have to switch from Alpine to Debian (for glibc), however.  A future update can slim this down significantly by only including just the runtime dependencies of the gosec and enry tools.

Additional included fixes:

* Engine: Fixed case where plugin registration on startup fails due to non-plugin directories (such as `__pycache__`).
* `Dockerfile.golang`: Fixed Docker style warning about `ENV` and prioritize our gosec/enry binaries in the PATH.
* `plugin_runner` utility: Simplified trimming the `$ECR` placeholder.

## Motivation and Context

Before, each container needed to make sure it had a compatible Python version as well as install the specific Python dependencies each plugin needed.  This was highly error-prone as different base images ship with different versions of Python, and we needed to manually update the dependency versions to match that of the engine.

A survey of the current container images shows that they ship with a wide variety of Python versions, and historically updating the dependency versions has been a source of runtime issues only caught late in the dev cycle.  In particular, attempting to make changes to the gosec plugin were stymied by the incompatible version combinations available in the base images.

## How Has This Been Tested?

## Types of changes

- [x] Documentation change

## Checklist


mdfleury-wbd (approved):
Exciting!

PR #300, opened 2024-10-08
Add Trufflehog to Heimdall Scans

g-marconet:
Add Trufflehog to Heimdall Scans

## Description

- Adds Trufflehog to Heimdall Scans with `depth=1`

## Motivation and Context

- Github Orgs have Github Advanced Security secrets scanning to intelligently scan for secrets
- Other VCS systems do not have an analogue to GHAS. Trufflehog can do that, so we enable it in Heimdall to get that coverage

## How Has This Been Tested?

Tests pass and working in dev environment

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/orchestrator/lambdas/repo_scan/repo_scan/repo_scan.py
+++ b/orchestrator/lambdas/repo_scan/repo_scan/repo_scan.py
@@ -26,10 +26,11 @@ warnings.filterwarnings("ignore", "No application metrics to publish*")
 
 ARTEMIS_API = os.environ.get("ARTEMIS_API")
 API_KEY_LOC = os.environ.get("ARTEMIS_API_KEY")
-DEFAULT_PLUGINS = ["gitsecrets", "base_images"]
+DEFAULT_PLUGINS = ["gitsecrets", "trufflehog", "base_images"]
 PROCESSED_MESSAGES = namedtuple("processed_messages", ["repos", "receipt_handles"])
 REPO_QUEUE = os.environ.get("REPO_QUEUE")
 REPO_DLQ = os.environ.get("REPO_DEAD_LETTER_QUEUE")
+SCAN_DEPTH = int(os.environ.get("SCAN_DEPTH", 1))
 SCAN_TABLE_NAME = os.environ.get("SCAN_TABLE") or ""
 
 log = Logger(service=APPLICATION, name="repo_scan")
@@ -190,6 +191,7 @@ def construct_repo_requests(repos: list) -> dict:
             "plugins": repo.get("plugins") or DEFAULT_PLUGINS,
             "batch_priority": True,
             "batch_id": repo.get("batch_id"),
+            "depth": SCAN_DEPTH,
         }
         metrics.add_metric(
             name="queued_repositories.count",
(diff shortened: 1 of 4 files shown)


orchestrator/lambdas/repo_scan/repo_scan/repo_scan.py
```diff
@@ -190,6 +191,7 @@ def construct_repo_requests(repos: list) -> dict:
             "plugins": repo.get("plugins") or DEFAULT_PLUGINS,
             "batch_priority": True,
             "batch_id": repo.get("batch_id"),
+            "depth": SCAN_DEPTH,
```
Kamsiy, line 194: Could we add a way to configure the scan depth?
Adding this parameter here could affect the scan depth used in other plugins

g-marconet: What did you have in mind?

It's grabbed from an environment variable [here](https://github.com/WarnerMedia/artemis/pull/300/files#diff-9f71be6ec05999d1c5bfe3dc81920c6950fccb69ef423f8f38819d88cac691b8R33), would you prefer configuring it a different way?

From a quick grep of the plugins directory, only Trufflehog (and Trufflehog Legacy) uses the depth. My understanding is that `depth` is a pretty specific flag intended (at the moment) for only the Trufflehog plugin.

If in the future we added other plugins that use `depth`, we should only do so where it's still referring to the number of commits to look backwards. In that case, it would still make sense for `depth` to be 1 to reduce Heimdall scan runtime.

Kamsiy: We could add a field to the `org-queue-lambda` payload that specifies the scan depth for each plugin?

Org-queue-lambda message format: https://github.com/WarnerMedia/artemis/tree/main/orchestrator/lambdas/org_queue#overview

Kamsiy: or we could hardcode scan_depth=1 to only apply to trufflehog

g-marconet: Could you expand on the value that would provide? I'm having trouble seeing it.

- Currently, Trufflehog is the only plugin that uses the `depth` parameter.
- I can't personally imagine a scenario where we would add another plugin that uses the `depth` parameter where we would want to control its depth separately, since the reason we're limiting it to 1 here is for performance reasons.

If we were to add configurable scan depth for each plugin, or even just handle Trufflehog alone, that would require a whole new request for each plugin and conditional logic that depends on the plugins themselves. That seems like a lot of complexity for not much benefit, IMO.

Kamsiy: That makes sense. I thought that scan_depth was used by all plugins. 

If its only applicable to trufflehog, it doesnt make sense to update it for all plugins.

Kamsiy (commented):
Everything looks good, 

Everything looks good 👍 

(nit)Please can you add the new environment variable to the `repo_scan` README https://github.com/WarnerMedia/artemis/tree/main/orchestrator/lambdas/repo_scan#environment-variables

Kamsiy (approved):

PR #310, opened 2024-10-14
ShellCheck plugin simplification and cleanups

ZoogieZork:
## Description

* Pass the `-S` (set minimum severity level) option to ShellCheck so we don't need to filter out findings ourselves.
* Use native Python to get the list of files rather than an `os.subprocess` call to `find`.
* Added unit test for `get_files()`.
* Added types.
* Handle the error case if the call to `shellcheck` fails.

## Motivation and Context

Cleans up several issues with the ShellCheck plugin noted over time.

## How Has This Been Tested?

Tested in nonprod environment.

## Types of changes

- [ ] Documentation change

## Checklist


backend/engine/plugins/shell_check/main.py
```diff
@@ -3,45 +3,38 @@
...
+import pathlib
 import subprocess
+from typing import Any, Optional
 
 from engine.plugins.lib import utils
 
 logger = utils.setup_logging("shell_check")
 
+# Filename extensions to include in scanning.
+INCLUDE_EXTS = {".sh", ".bash", ".ksh", ".bashrc", ".bash_login", ".bash_logout", ".bash_profile", ".bash"}
```
g-marconet, line 15: Nit: `.bash` is repeated

g-marconet (commented):
Looks good, just one small non-functional thing

backend/engine/plugins/shell_check/main.py
ZoogieZork: Good catch!  Fixed.

g-marconet (approved):
LGTM

PR #317, opened 2024-10-22
checking for non existent keys

mdfleury-wbd:
Testing for non-existent keys so plugin doesn't throw errors

## How Has This Been Tested?
Tested in nonprod

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/backend/engine/plugins/node_dependencies/parse.py
+++ b/backend/engine/plugins/node_dependencies/parse.py
@@ -119,7 +119,7 @@ def _find_versions_v2(lockfile: dict, nodes: list, resolver: LineNumberResolver)
                 "integrity": pkg.get("integrity", ""),
                 "dev": pkg.get("dev", False),
                 "filename": source.get("filename", ""),
-                "line": source.get("line", ""),
+                "line": source.get("line", 0),
             }
         )
     return versions
--- a/backend/engine/plugins/node_dependencies/parse.py
+++ b/backend/engine/plugins/node_dependencies/parse.py
@@ -111,15 +111,15 @@ def _find_versions_v2(lockfile: dict, nodes: list, resolver: LineNumberResolver)
     for node in nodes:
         pkg = lockfile["packages"][node]
         path = f"/{node}".replace("/node_modules/", ">")[1:]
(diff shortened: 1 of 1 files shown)


backend/engine/plugins/node_dependencies/parse.py
```diff
@@ -111,15 +111,15 @@ def _find_versions_v2(lockfile: dict, nodes: list, resolver: LineNumberResolver)
...
         pkg = lockfile["packages"][node]
         path = f"/{node}".replace("/node_modules/", ">")[1:]
-        source = resolver.find(pkg["integrity"]) or resolver.find(path)
+        source = resolver.find(pkg.get("integrity")) or resolver.find(path)
         versions.append(
             {
                 "path": path,
                 "version": pkg["version"],
-                "integrity": pkg["integrity"],
+                "integrity": pkg.get("integrity", ""),
```
g-marconet, line 119: Is empty string the best value for empty here? How would the rest of Artemis handle empty string as opposed to something like "None provided"?

(Not suggesting we change it here, just trying to understand. This also applies to `filename` below)

backend/engine/plugins/node_dependencies/parse.py
```diff
@@ -111,15 +111,15 @@ def _find_versions_v2(lockfile: dict, nodes: list, resolver: LineNumberResolver)
...
                 "dev": pkg.get("dev", False),
-                "filename": source["filename"],
-                "line": source["line"],
+                "filename": source.get("filename", ""),
+                "line": source.get("line", ""),
```
g-marconet, line 122: What type is `line` when it does exist in `source`? If it's a number, does it make sense to sometimes have it be a string?

mdfleury-wbd: Yeah, it makes more sense as a string. I'll fix that.

backend/engine/plugins/node_dependencies/parse.py
mdfleury-wbd: If you go up in this file, we never do anything with the integrity field, so it shouldn't change anything.

PR #319, opened 2024-10-22
Update Checkov and CFN-Lint

g-marconet:
Update Checkov and CFN-Lint

## Description

- Updates Checkov to `3.2.268`
  - Packaging gets bumped to `23.2`, which is what Checkov requires
- CFN-Lint gets bumped to `1.18.1`
  - The latest version removes a conflict with Checkov on `jsonschema` versions
- Also removes some things we were doing to accommodate the old Checkov version, like pinning `setuptools` and `virtualenv` versions in the Github Actions Workflows

## Motivation and Context

- Checkov likes to pin dependencies, which made the old version kinda problematic as its dependencies were conflicting with other things.
- CFN-Lint needed to be updated in order to update Checkov

## How Has This Been Tested?

- Tests pass
- Working in dev environment

## Types of changes

- [ ] Documentation change

## Checklist


davakos (approved):

backend/engine/tests/test_plugin_checkov.py
ZoogieZork, line 1: For a follow-up PR, we should be able to un-comment the "New findings in Checkov 3.2." block.  Unless something changed since the last time we tried to upgrade Checkov, those extra two test cases should still be valid. :)

PR #323, opened 2024-10-23
Clean up bundler-audit plugin

ZoogieZork:
## Description

* Fix error handling to handle errors other than "No such file or directory".
* Ensure that the output from bundler-audit is not colorized.
* Fix type hints and docstrings.  Added additional documentation and test cases.

## Motivation and Context

Fixes numerous issues discovered while testing switching to the "boxed" runner for Ruby-oriented plugins.

## How Has This Been Tested?

Tested in non-prod environment.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation change

## Checklist

- [x] My code follows conforms to the coding standards.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #338, opened 2024-10-29
Overhaul handling of errors from bundler-audit

ZoogieZork:
## Description

Parsing of stdout and stderr from running bundler-audit is now more consistent, with better detection of error conditions.

A [previous fix](https://github.com/WarnerMedia/artemis/pull/323) adjusted the stderr parsing so that it no longer ignores/mangles error messages, but this exposed the issue that git (which is called by bundler-audit) writes its progress messages to stderr, which was being captured as errors.  Additionally, in one case the tool emits a fatal error message to stdout instead of stderr.

We now check the exit code of the tool as well as if any findings are present in order to determine if the errors should be reported.

A number of test cases were added based on actual test runs of bundler-audit so that the tests better reflect reality.

## Motivation and Context

Fix error messages mistakenly being reported for bundler-audit runs.

## How Has This Been Tested?

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/backend/engine/plugins/bundler_audit/main.py
+++ b/backend/engine/plugins/bundler_audit/main.py
@@ -2,11 +2,12 @@
 Bundler-audit plugin
 """
 
+from dataclasses import dataclass, field
 import glob
 import json
 import os
 import subprocess
-from dataclasses import dataclass, field
+from typing import Optional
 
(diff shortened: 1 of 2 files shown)


g-marconet (approved):
LGTM

PR #339, opened 2024-10-29
Support more NPM license conventions

g-marconet:
- Extends the NPM license retriever to cover more cases

## Description

- Can now handle SPDX license expressions
  - Which is the currently-supported way to have multiple licenses: https://docs.npmjs.com/cli/v10/configuring-npm/package-json#license
  - We opted not to lose the AND/OR/WITH logic in this pull request by extracting all the mentioned license versions. SPDX expressions should be preserved as-is
    - Another option is to parse the expression and return all licenses mentioned, but then we'd lose the AND/OR/WITH relationship. (We are not doing this)
- Also handles more "deprecated" license cases
  - Like `license: { type: str }`
  - Like `licenses: [ { type: str } ]`

## Motivation and Context

There were common license patterns that we were not reading

## How Has This Been Tested?

Tests pass and working in dev environment

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/backend/lambdas/maintenance/license_retriever/license_retriever/retrievers/npm.py
+++ b/backend/lambdas/maintenance/license_retriever/license_retriever/retrievers/npm.py
@@ -22,12 +22,13 @@ def retrieve_npm_licenses(name: str, version: str) -> list:
 
     for item in license:
         item_type = type(item)
+
         if item_type is str:
             result.append(item.lower())
         elif item_type is dict and 'type' in item:
             result.append(item['type'].lower())
         else:
-            LOG.error(f'Unexpected license format for npm package, "{name}@{version}". Result was: {json.dumps(item)}')
+            LOG.error(f'Unexpected license format for npm package, "{name}@{version}". License item was: {json.dumps(item)}')
 
     return result
 
--- a/backend/lambdas/maintenance/license_retriever/license_retriever/retrievers/npm.py
+++ b/backend/lambdas/maintenance/license_retriever/license_retriever/retrievers/npm.py
@@ -1,7 +1,6 @@
 from time import sleep
 from typing import Optional, Union
 
-import json
 import requests
 
 from artemislib.logging import Logger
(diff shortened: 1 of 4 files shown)


backend/lambdas/maintenance/license_retriever/license_retriever/retrievers/npm.py
```diff
@@ -38,3 +30,62 @@ def get_package_info(name: str, version: str) -> dict:
...
+
+
+def get_license(item: Union[dict, str], package_name: str) -> Optional[str]:
+    if type(item) is str:
+        return item.lower()
+    elif type(item) is dict:
+        if "type" in item and type(item["type"]) is str:
+            return item["type"].lower()
+
+    LOG.warning('Could not parse a license for "%s". It may be unconventional.', package_name)
```
g-marconet, line 90: Is warning the right log level for this?

Essentially, I wanted some log to point out if conventions change or if there is a common convention we miss. However, this input is untrusted, so an NPM package messing up the convention or similar will also trigger this path.

backend/lambdas/maintenance/license_retriever/license_retriever/retrievers/npm.py
```diff
@@ -38,3 +30,62 @@ def get_package_info(name: str, version: str) -> dict:
...
+    #     { type: "GPL-3.0", url: "..." },
+    #     { type: "MIT", url: "..." }
+    # ]
+
+    if "license" in package_info:
+        license = package_info["license"]
+    elif "licenses" in package_info:
+        license = package_info["licenses"]
+    else:
+        LOG.debug('No license information in package info for "%s"', package_name)
```
g-marconet, line 61: For reviewer, should this be debug? We log as "Error" on L31 when we can't find the package info.

backend/lambdas/maintenance/license_retriever/license_retriever/retrievers/npm.py
Kamsiy: A warning is perfect here

backend/lambdas/maintenance/license_retriever/license_retriever/retrievers/npm.py
Kamsiy: I think the information logged here should be a `warning` . But it also depends on how common this case is. 
If its really common, it makes sense to have it a `debug` log. If its not, then it should be a `warning`

Kamsiy (approved):

PR #344, opened 2024-11-05
Add CICD Tool Discovery Plugin to Backend

g-marconet:
Adds a CICD Tool Discovery plugin to the backend

## Description

- Makes a best effort to identify CICD tools based on config file conventions.
- Currently supports:
  - AWS CodeBuild
  - BitRise
  - CircleCI
  - Electron Forge
  - Github Actions
  - Gitlab CI
  - Jenkins
  - Teamcity
  - Travis CI

## Motivation and Context

- Knowledge of CICD tool used by a given repository can inform security decisions

## How Has This Been Tested?

- Working in dev environment

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/backend/engine/plugins/cicd_tools/detectors/aws_codebuild_detector.py
+++ b/backend/engine/plugins/cicd_tools/detectors/aws_codebuild_detector.py
@@ -32,11 +32,16 @@ class AWSCodeBuildDetector(Detector):
             "errors": [],
         }
 
-        buildspec_dirs = Path(path).rglob("**/*buildspec*/")
-        buildspec_files = Path(path).rglob("**/*buildspec*.yml")
+        base = Path(path)
+
+        buildspec_dirs = base.rglob("**/*buildspec*/")
+
+        buildspec_files = list(base.rglob("**/*buildspec*.yml"))
+        buildspec_files.extend(base.rglob("**/*buildspec*.yaml"))
 
         for dir in buildspec_dirs:
-            yaml_files = dir.rglob("**/*.yml")
+            yaml_files = list(dir.rglob("**/*.yml"))
+            yaml_files.extend(dir.rglob("**/*.yaml"))
 
             for file in yaml_files:
                 result["in_use"] = True
--- a/backend/engine/plugins/cicd_tools/detectors/aws_codebuild_detector.py
+++ b/backend/engine/plugins/cicd_tools/detectors/aws_codebuild_detector.py
@@ -1,0 +1,49 @@
+import json
+
+from pathlib import Path
+from typing import Callable
+
+from engine.plugins.cicd_tools.interfaces.detector import Detector, DetectorResult
+
+
+ValidatorFn = Callable[[Path], bool]
+
+ID = "aws_codebuild"
+NAME = "AWS CodeBuild"
+
+
+class AWSCodeBuildDetector(Detector):
+    """
+    AWS CodeBuild has "buildspec.yml" as a convention, but it's customizable.
+    We have noticed any of the following patterns, so they should all be reported as CodeBuild:
+    - buildspec.yml
+    - buildspec-deploy.yml
+    - /buildspec/region/deploy.yml
+    """
+
+    def check(self, path: str) -> DetectorResult:
+        result: DetectorResult = {
+            "id": ID,
+            "name": NAME,
+            "configs": [],
+            "in_use": False,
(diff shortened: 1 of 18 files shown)


backend/engine/plugins/cicd_tools/interfaces/detector.py
```diff
@@ -0,0 +1,17 @@
...
+    in_use: bool
+
+    debug: list[str]
+    alerts: list[str]
+    errors: list[str]
+
+
+class Detector:
+    def check(self, path: str) -> DetectorResult:
+        raise NotImplementedError()
```
ZoogieZork, line 17: (minor) This is a good opportunity to use [`ABC`](https://docs.python.org/3/library/abc.html#abc.abstractmethod) and [`@abstractmethod`](https://docs.python.org/3/library/abc.html#abc.abstractmethod). If we add additional required methods in the future, the type-checker will warn us about missing methods.

```python
from abc import ABC, abstractmethod

class Detector(ABC):
    @abstractmethod
    def check(self, path: str) -> DetectorResult: ...
```

Example:

```python
class Foo(Detector):
    ...

foo = Foo()  # Cannot instantiate abstract class "Foo": "Detector.check" is not implemented
```

ZoogieZork (dismissed):

backend/engine/plugins/cicd_tools/detectors/aws_codebuild_detector.py
```diff
@@ -0,0 +1,49 @@
...
+            "name": NAME,
+            "configs": [],
+            "in_use": False,
+            "debug": [],
+            "alerts": [],
+            "errors": [],
+        }
+
+        buildspec_dirs = Path(path).rglob("**/*buildspec*/")
+        buildspec_files = Path(path).rglob("**/*buildspec*.yml")
```
davakos, line 36: this question goes for all instances - should we support both `.yml` and `.yaml` file suffixes as I see them used pretty interchangeably?

backend/engine/plugins/cicd_tools/main.py
```diff
@@ -0,0 +1,153 @@
...
+    )
+
+
+def find_tools(path: str) -> list[DetectorResult]:
+    return [detector.check(path) for detector in detectors]
+
+
+def get_details(results: list[DetectorResult]) -> dict[str, dict[str, CICDToolsDetails]]:
+    return {
+        "cicd_tools": {
```
davakos, line 129: I know this is the JSON output I proposed for this plugin, but this was just a swag - please review just to ensure that it complies with output from similar APIs in the system... for example, `scan_options.plugins` in a scan object is just just an array of the internal plugin ids (no display name), and we maintain an internal mapping in backend & UI to the display names. Would this be a better and more consistent approach?

backend/lambdas/api/spec.yaml
```diff
@@ -2079,6 +2079,7 @@ components:
               - php_sensio_security_checker
               - python_code_checker
               - technology_discovery
+              - cicd_tools
```
davakos, line 2082: do we also need to update the `artemis-scan.sh` script to add this plugin?

backend/engine/plugins/cicd_tools/detectors/aws_codebuild_detector.py
g-marconet: Good catch, added

backend/engine/plugins/cicd_tools/main.py
g-marconet: I don't think a `scan_options.plugins` type mapping would make sense in this context, but I also don't have a good grasp of what you are envisioning there. At a minimum, it would have the added complexity of causing us to have to edit things in two places when we want to add additional CICD tools or modify the current ones.

The current pattern is similar to the one in the repo healthcheck plugins, where there's an ID for the particular finding and then the plugins also provide a plaintext `display_name` analogue that eventually gets displayed in the UI. It keeps everything localized in the plugin, so changing things is more intuitive, IMO.

backend/lambdas/api/spec.yaml
g-marconet: Discussed this offline. For posterity, `artemis-scan.sh` now includes all categories by default and has an explicit block list of plugins not to run by default, so it does not need to be updated.

davakos (approved):

PR #352, opened 2024-11-12
Mount source volume as read-only by default

ZoogieZork:
## Description

Mount the source code volume in plugin containers as read-only by default.  Individual plugins which need to write to the source tree (usually to generate lockfiles) can indicate so in settings.json.

We make read-only the default since the majority of plugins do not need to write to the source tree.  After reviewing the plugins, only eight write to the source tree.

## Motivation and Context

Mounting the source volume as read-only means that we can skip the clean-up step, avoids potential data / permissions issues during the clean-up.  In the future, this will allow us to run these plugins in parallel.

## How Has This Been Tested?

## Types of changes

- [x] Documentation change

## Checklist


mdfleury-wbd (approved):
Love it

PR #363, opened 2024-11-18
Provide a temporary "scratch" volume to plugins

ZoogieZork:
## Description

Plugins are now provided with a temporary named volume which is automatically deleted after the plugin runs. The name of this volume is passed to the plugin via the engine var `temp_vol_name`.

This sets the framework for plugins which, in turn, execute a tool via a container (i.e. `docker run`).  These plugins can mount this named volume into the sub-container in order to pass files back and forth.  For example, the plugin can prepare config files and pass them to the sub-container, then receive result files written by the sub-container.

* Docker SDK added to dependencies, kicking off migrating away from running the Docker CLI via `subprocess.run()`.
* Renamed the engine's `docker` package to `oci` to side-step the name conflict with the Docker SDK.
* The Plugin Runner utility now mounts the target directory in the engine container, which is then inherited by the plugin container. This more closely matches (not perfectly, but closer) the deployed engine.
* Invalid JSON files passed to the Plugin Runner (e.g. `engine-vars.json`) are now treated as an error instead of a warning. This is because we need to merge the `temp_vol_name` and `engine_id` settings into the provided JSON.  If there's an _actual_ need to test passing invalid JSON to the plugin, it can be done via the debug shell. :)

## Motivation and Context

Support for plugins which execute tools via sub-containers.

## How Has This Been Tested?

## Types of changes

- [x] Documentation change

## Checklist


mdfleury-wbd (approved):
I don't like the oci name 🤷

ZoogieZork:
> I don't like the oci name 🤷

Yeah, my reasoning is that the module is mostly concerned with building [OCI images](https://opencontainers.org/), which isn't necessarily specific to Docker (i.e., Docker in this case is just an implementation detail).   We can still rename it later if we come up with a better name, it just needed to not conflict with `import docker`.

PR #365, opened 2024-11-20
Add "authToken" auth type for .npmrc

g-marconet:
Adds the option of using authToken for authenticating with .npmrc

## Description

- If an `authToken` property exists in a scope, we will use it with `_authToken` rather than the current `username` and `_password` pattern
- This is technically breaking, but it would require a scope to already include an `authToken` property despite it previously having no value.

## Motivation and Context

`authToken` is preferred by some repository managers, so we should support it.

## How Has This Been Tested?

- Working in dev environment

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/backend/engine/plugins/lib/write_npmrc.py
+++ b/backend/engine/plugins/lib/write_npmrc.py
@@ -98,11 +98,10 @@ def write_npmrc(logger, npmrc, scope_list: list) -> None:
     with open(npmrc, "a") as f:
         for scope in scope_list:
             logger.info(f"Writing {scope['scope']} config to %s", npmrc)
-            
+
             if "authToken" in scope:
                 npm_config = build_npm_auth_token_config(**scope)
             else:
                 npm_config = build_npm_config(**scope)
-            
+
             f.write(npm_config)
-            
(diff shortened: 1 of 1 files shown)


backend/engine/plugins/lib/write_npmrc.py
```diff
@@ -79,9 +79,29 @@ def build_npm_config(scope, registry, token, username, email, **kwargs):
...
+    """
+    gets the variables necessary to return the npmrc config with _authToken
+    This provides npm the necessary info to get packages from private sources.
+    **kwargs exists at the end to pick up anything extra in the config dictionary that we dont need.
+    """
+    return (
+        f"@{scope}:registry=https://{registry}\n"
+        f"//{registry}:_authToken={authToken}\n"
+        f"//{registry}:email={email}\n"
+        f"//{registry}:always-auth=true\n"
```
ZoogieZork, line 92: I'm not intimately familiar with the .npmrc file format -- do we need to escape any special characters here?

g-marconet: If the input were untrusted, yes. Injection would be possible.

That said, the scopes are pulled down from AWS Secrets Manager in the same file. This should never see untrusted input (and, FWIW, the current `build_npm_config` logic does no escaping, either).

ZoogieZork (approved):

PR #366, opened 2024-11-20
Run Checkov via the Docker image

ZoogieZork:
## Description

Runs Checkov via the [Docker image](https://hub.docker.com/r/bridgecrew/checkov) instead of installed via pip/pipenv.

Additionally, cleaned up lint and typehint warnings, as well as improved the comments.

> [!NOTE]
> Checkov will be removed from the Pipfile as part of the next dependency update run.

## Motivation and Context

Checkov has a very large and picky list of dependencies, which been a major source of dependency conflicts.  By using the Docker image instead, this allows us to keep other dependencies updated while updating Checkov regularly.

## How Has This Been Tested?

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/backend/engine/plugins/checkov/main.py
+++ b/backend/engine/plugins/checkov/main.py
@@ -117,7 +117,7 @@ def run_checkov(
         shutil.copytree(path, srcdir)
         checkov_command += ["-d", "/tmp/base/work"]
 
-    LOG.info(f"Starting Checkov in container, path: {path}")
+    LOG.info(f"Starting {CHECKOV_IMG_REF} in container, path: {path}")
 
     stderr = docker_client.containers.run(
         CHECKOV_IMG_REF,
--- a/backend/engine/plugins/checkov/main.py
+++ b/backend/engine/plugins/checkov/main.py
@@ -5,7 +5,6 @@ Checkov Plugin
 import docker
 import json
 import os
-import shutil
 import subprocess
 from typing import Optional, TypedDict, Union
 from os.path import abspath
@@ -38,8 +37,6 @@ def main():
 
     errors: list[str] = []
 
-    engine_id: Optional[str] = args.engine_vars["engine_id"]
-
     (temp_vol_name, temp_vol_mount) = str(args.engine_vars.get("temp_vol_name", "")).split(":")
     if not temp_vol_name or not temp_vol_mount:
         errors.append("Temporary volume not provided")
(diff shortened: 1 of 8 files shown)


backend/engine/plugins/checkov/main.py
```diff
@@ -21,29 +34,60 @@ def main():
...
+
+    errors: list[str] = []
+
+    (temp_vol_name, temp_vol_mount) = str(args.engine_vars.get("temp_vol_name", "")).split(":")
+    if not temp_vol_name or not temp_vol_mount:
+        errors.append("Temporary volume not provided")
+
+    (working_src, working_mount) = str(args.engine_vars.get("working_mount", "")).split(":")
+    if not working_src or not working_mount:
+        errors.append("Working volume not provided")
```
g-marconet, line 46: How are `temp_vol_name` and `working_mount` defined? Do we control them?

ZoogieZork: These are set here: https://github.com/WarnerMedia/artemis/blob/zoogiezork/checkov-docker-2/backend/engine/utils/plugin.py#L167-L168

backend/engine/plugins/checkov/main.py
```diff
@@ -144,10 +205,12 @@ def parse_checkov(checkov_output: list[dict], repo_path: str, config_dir: str, c
     return (findings, error)
 
 
-def get_config_dir(repo_path: str, config: dict) -> tuple[str, str]:
+def get_config_dir(config: dict) -> tuple[Path, Optional[str]]:
     """
     Determine if config directory should be from S3 or default (local)
     """
+    # TODO: Write to the temporary volume instead of the plugin source directory.
```
g-marconet, line 212: Nit: TODO possibly left in? Would this be better as an issue in an issue tracker? (also L222 below)

g-marconet (approved):
LGTM

backend/engine/plugins/checkov/main.py
ZoogieZork: Yep, this is intentional -- it's a note for the next PR in this series (or if I don't get around to it, will be added to the backlog).

The existing implementation of this writes the retrieved configuration into the actual plugin source directory, which persists between runs.  This *technically* works, so doesn't block this PR, but is definitely not good and potentially cause issues in some corner cases.

As far as I can tell, we're not actually using this "feature" so it's a lower priority to fix.

PR #370, opened 2024-11-27
Gracefully handle Checkov plugin shutdown

ZoogieZork:
## Description

Installs a signal handler to gracefully terminate the Checkov container.

Without this, when the plugin is terminated the Checkov container would continue running and then wouldn't be removed afterwards.

> [!NOTE]
> This currently only has an effect when running the Checkov plugin via the Plugin Runner utility.
> The engine currently sends a SIGKILL to the plugin when the timeout expires, which we can't trap.

## Motivation and Context

Properly clean up the container when running via the Plugin Runner utility, and also prepare the plugin for when we add graceful shutdown to the engine.

## How Has This Been Tested?

Tested in non-prod environment as well as the Plugin Runner.

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/backend/engine/plugins/checkov/main.py
+++ b/backend/engine/plugins/checkov/main.py
@@ -4,9 +4,12 @@ Checkov Plugin
 
 import boto3
 import docker
+import docker.errors
 import json
 import os
 import shutil
+import signal
+import sys
 from typing import Optional, TypedDict, Union
 from os.path import abspath
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


backend/engine/plugins/checkov/main.py
```diff
@@ -116,8 +149,9 @@ def run_checkov(
 
     stderr = docker_client.containers.run(
         CHECKOV_IMG_REF,
+        name=container_name,
         command=checkov_command,
-        remove=True,
+        auto_remove=True,
```
ZoogieZork, line 154: `auto_remove` is preferred since the Docker daemon will handle removing the stopped container on exit.  `remove` relies on the client to remove the container, which it won't have a chance to do if the plugin is aborting.

g-marconet (approved):
LGTM

PR #371, opened 2024-11-27
Add Plugin Runner Toolbox

ZoogieZork:
## Description

The Plugin Runner now attaches a "toolbox" container to aid in debugging plugins:

* **Provide BusyBox shell and command-line tools.** This enables support for containers which do not have a built-in shell.
* **Colorize stderr.** We can now disambiguate between stderr and stdout from the plugin. We also make sure that stdout and stderr lines don't write over each other.
* **Basic plugin result linting.** Validate the plugin JSON output for issues.

The plugin result linter doesn't yet validate the objects in `details` since the schema is determined by the plugin type, which the linter doesn't know (yet).

The fields which are considered as required is based on the plugin documentation here: [README.md](https://github.com/WarnerMedia/artemis/blob/760e1864a6b8dce0944b18bdfb3e4b2d0bca6960/backend/engine/plugins/README.md#mainpy)
We may adjust these rules later in conjunction with updating the documentation.

### Screenshot

image

## Motivation and Context

I've been doing a lot of debugging of plugins lately.

## How Has This Been Tested?

Tested locally.

## Types of changes

- [ ] Documentation change

## Checklist


backend/utilities/plugin_runner/toolbox/lint.go
```diff
@@ -0,0 +1,55 @@
...
+	errs := jsonValidator.Validate(buf, &result)
+	retv = append(retv, errs...)
+
+	if result.Truncated != nil && *result.Truncated {
+		retv = append(retv, jsonValidator.ValidationError{
+			Field: "truncated", Message: "Must be false",
+		})
+	}
+
+	//TODO: Validate details based on plugin type.
```
g-marconet, line 52: Nit: Just want to call attention to the TODO to make sure it's tracked

ZoogieZork: Yep, this will be in a follow-up PR.  The plugin.sh script will need to look up and pass the plugin type to the toolbox runner so the linter knows what schema to expect.

g-marconet (approved):
LGTM

PR #375, opened 2024-12-02
Parallelize backend unit tests

ZoogieZork:
## Description

This is only enabled by default in CI runs since the output is harder to follow, but anybody can enable it locally using the same `PYTEST_EXTRA_ARGS` technique shown here.

```bash
PYTEST_EXTRA_ARGS='-n auto --dist worksteal'
```

Testing in CI shows that it typically reduces the time it takes to the run the backend unit tests from around a minute to 30-50 secs, depending on how the tests are distributed between the worker processes.  Even in the worst case, this should always be an improvement.

## Motivation and Context

Shorten the time waiting for backend unit tests to complete.

## How Has This Been Tested?

Tested locally (with up to `-n 24`) and in CI.

## Types of changes

- [ ] Documentation change

## Checklist

- [] My change requires a change to the documentation.


mdfleury-wbd (approved):
Makes it look so easy

PR #376, opened 2024-12-03
Generate CWE mapping directly

ZoogieZork:
## Description

Adds a utility script to generate a JSON mapping of CWE IDs to names.

This generated file replaces the [cwe2](https://pypi.org/project/cwe2/) package.  The cwe2 package weighs in at 45 MB, making it one of the largest dependencies we pull in, but we only need the ID-to-name mapping, which is a 56 KB JSON file.

## Motivation and Context

Reduces the size of the backend dependencies, which in turn reduces the size of the common lambda layer and boxed plugin package.

## How Has This Been Tested?

TODO

## Types of changes

- [ ] Documentation change

## Checklist


backend/utilities/cwe_map_generator/main.py
```diff
@@ -0,0 +1,111 @@
...
+    with requests.get(src, headers=headers, stream=True) as req:
+        if req.status_code == 304:
+            LOG.info("Database up-to-date.")
+            return
+        req.raise_for_status()
+
+        with tempfile.TemporaryDirectory() as tempdir:
+            outfile = Path(tempdir) / "download.tmp"
+            with outfile.open("wb") as f:
+                for chunk in req.iter_content(chunk_size=64 * 1024):
```
breedenc, line 51: Possibly a naive question, is there a special significance to this chunk size of 64 KiB that we should document in code comments?

ZoogieZork: Nope, just an arbitrary buffer size.

breedenc (approved):

PR #380, opened 2024-12-05
Re-enable default caching for docker build

ZoogieZork:
## Description

When running `docker build`, removes the `--no-cache` and `--force-rm` options from always being set.

If needed, these can be set using the `DOCKER_BUILD_EXTRA_ARGS` var, e.g.:

```
make dist/docker/engine DOCKER_BUILD_EXTRA_ARGS='--no-cache --force-rm'
```

This will not affect CI deploys since we assume that each CI run will start with an empty cache anyway.

To inaugurate this change, Dockerfile.engine has been updated to use cache mounts for Apt, Pip, and Pipenv.

## Motivation and Context

We want to make use of both layer caches and cache mounts when building our container images to improve the local dev cycle.  There may be additional tweaks needed to .dockerignore to avoid unnecessary rebuilds, but this is a first start.

## How Has This Been Tested?

Tested in non-prod environment as part of the 2024-12-09 "omnibus" deployment with Django 4.2.

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/backend/Dockerfiles/Dockerfile.engine
+++ b/backend/Dockerfiles/Dockerfile.engine
@@ -1,3 +1,4 @@
+# syntax=docker/dockerfile:1
 ARG ENGINE_BASE=3.9-slim-bullseye
 
 ###############################################################################
@@ -7,7 +8,8 @@ FROM python:3.9-bullseye AS builder
 
 # Install the Docker repository
 # hadolint ignore=DL3008
-RUN apt-get update \
+RUN --mount=type=cache,target=/var/cache/apt \
+    apt-get update \
     && apt-get install -y --no-install-recommends \
         lsb-release \
         software-properties-common \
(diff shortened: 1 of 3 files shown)


ZoogieZork:
.unlock nonprod

backend/Dockerfiles/Dockerfile.engine
```diff
@@ -17,19 +19,20 @@ RUN apt-get update \
 SHELL ["/bin/bash", "-o", "pipefail", "-c"]
 RUN wget --quiet -O - https://download.docker.com/linux/debian/gpg | apt-key add -
 
-RUN pip install --no-cache-dir \
+# hadolint ignore=DL3042
+RUN --mount=type=cache,target=/root/.cache/pip \
+    pip install \
     'packaged==0.6.0' \
-    'pipenv~=2024.0' \
-    'setuptools<74' \
```
breedenc, line 23: I don't see these changes mentioned in the PR description, are these pins no longer required due to issues that have been resolved?

backend/Dockerfiles/Dockerfile.engine
```diff
@@ -7,7 +8,8 @@ FROM python:3.9-bullseye AS builder
 
 # Install the Docker repository
 # hadolint ignore=DL3008
-RUN apt-get update \
+RUN --mount=type=cache,target=/var/cache/apt \
```
breedenc, line 11: Would it be a best practice to specify `sharing=locked` for the Apt cache mount, even if we don't expect the cache to ever be shared between parallel builds?

I ask because it's mentioned here: https://docs.docker.com/build/cache/optimize/#use-cache-mounts

backend/Dockerfiles/Dockerfile.engine
ZoogieZork: Good catch.  Yep, when we upgraded Checkov a while back, we didn't need these version pins anymore but missed removing them from Dockerfile.engine.

backend/Dockerfiles/Dockerfile.engine
ZoogieZork: Good catch, yep, we definitely want to use `sharing=locked` here since long term we want to be able to parallelize our builds (like we currently do with the Java images).

ZoogieZork: I've set `sharing=locked` on both Apt (since it's documented as necessary) and Pipenv (since I can't find any info on either way, so setting it to be safe).

backend/Dockerfiles/Dockerfile.engine
breedenc: Linking here for convenience: #319

breedenc (approved):

PR #381, opened 2024-12-07
Dependency Check plugin cleanups

ZoogieZork:
## Description

Various cleanups and fixes to the OWASP Dependency Check plugin:

* Cleaned up lint and typecheck warnings.
* The results JSON can be quite large, so we stream the JSON conversion directly to stdout rather than stringifying it first.
* Write the output of the Dependency Check tool to a temporary directory instead of the source tree.  This is part of the journey to be able to point the source tree as read-only for future safety and parallelism.

## Motivation and Context

Cherry-picked cleanups from a separate larger branch.

## How Has This Been Tested?

Tested in non-prod environment as part of the 2024-12-09 "omnibus" deployment with Django 4.2.

## Types of changes

- [ ] Documentation change

## Checklist


backend/engine/plugins/owasp_dependency_check/main.py
```diff
@@ -118,13 +123,13 @@ def parse_scanner_output_json(repo_path: str, returncode: int) -> dict:
...
-        data = json.load(json_file)
-        results = parse_vulnerabilities(data)
-        errors = parse_errors(data)
+    try:
+        with open(report_path) as json_file:
+            data = json.load(json_file)
+            results = parse_vulnerabilities(data)
+            errors = parse_errors(data)
+    except FileNotFoundError:
+        return {"output": [], "errors": [f"No report file found: {report_path}"], "success": success}
```
breedenc, line 132: Is there a non-error state where the report would not be found, or does Dependency Check write the report file every time? If we expect it to write a report file every time it runs, should `success` here be `False` instead of based on Dependency Check's return code? 

Or is this not desirable because the lack of report is surfaced in `errors` and we want `success` to reflect Dependency Check's exit code?

ZoogieZork: Good questions. This handling was retained from the original code (albeit with a more descriptive error message), it was moved from being a separate check into handling the exception.

Since we don't know what cases this might occur (and future Dependency Check versions may change this behavior), I think the best case here would be to treat this as an internal error, i.e., return `success=True` so we don't fail the CI build, but set the error so it can be investigated.

breedenc (dismissed):

backend/engine/plugins/owasp_dependency_check/main.py
ZoogieZork: Changes made and added a comment about it.

breedenc (approved):

PR #389, opened 2024-12-12
Add "cicd_tool" filter to repository search

g-marconet:
Adds a "cicd_tool" filter to the repository search endpoint

## Description

- Adds a "cicd_tool" filter to the repository search endpoint
- This is only in the API for now. We may have to follow up with further changes if this proves not performant enough

## Motivation and Context

- We have CICD Tool data and would like to search for it

## How Has This Been Tested?

- Working in dev environment

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/backend/lambdas/api/search_repositories/search_repositories/get.py
+++ b/backend/lambdas/api/search_repositories/search_repositories/get.py
@@ -1,7 +1,5 @@
-from django.db.models.query import QuerySet
-
 from artemisapi.const import SearchRepositoriesAPIIdentifier
-from artemisdb.artemisdb.models import Repo, Scan
+from artemisdb.artemisdb.models import Repo
 from artemisdb.artemisdb.paging import Filter, FilterMap, FilterMapItem, FilterType, PageInfo, apply_filters, page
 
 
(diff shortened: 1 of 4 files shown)


backend/lambdas/api/search_repositories/search_repositories/get.py
```diff
@@ -5,6 +5,9 @@
 from artemisdb.artemisdb.paging import Filter, FilterMap, FilterMapItem, FilterType, PageInfo, apply_filters, page
 
 
+CICD_TOOL_PARAM = "cicd_tool"
```
breedenc, line 8: Would it make sense to store this in either `artemisapi.const` or `artemisdb.artemisdb.consts` since the same string is declared here and in `paging.py` (although it's in a list in the latter)?

backend/libs/artemisdb/artemisdb/artemisdb/models.py
```diff
@@ -233,6 +233,13 @@ def in_scope(cls, scopes: list[list[list[str]]]) -> models.QuerySet:
             # No filter was built up (no scopes?) so return nothing
             return cls.objects.none()
 
+    @classmethod
```
breedenc, line 236: Should this be a static method since `cls` isn't used/needed?

g-marconet: Good catch, but I think it's the other way around. We should use `cls` instead of `Repo` to match the pattern in `in_scope` above

breedenc (approved):

PR #394, opened 2024-12-17
Plugin Runner: Schema validation for findings

ZoogieZork:
## Description

Adds schema validation for the `details` field of plugin results.

The validation is now specified using JSON schemas. The results are somewhat less user-friendly, but in the future we may be able to sync it with the API spec.

Only the three finding types which are documented in [plugins/README.md](https://github.com/WarnerMedia/artemis/blob/2f4c4b5a5916a07ed372b39c8ff3719744c3d92f/backend/engine/plugins/README.md) are included:

* Secrets
* Static Analysis
* Vulnerability

As we fill-out our documentation of the plugin types, we will update the schemas as well.

Example output:

image

## Motivation and Context

Enable debugging of plugin issues with the format of the findings which have caused runtime issues.

## How Has This Been Tested?

Tested locally.

## Types of changes

- [ ] Documentation change

## Checklist


mdfleury-wbd (approved):
I haven't written enough Go to comment much there, but everything makes sense to me. Would it make sense to eventually convert plugin.sh to something else as that file is getting more complicated?

PR #396, opened 2024-12-18
Add CICD Tools Plugin to UI

g-marconet:
Adds CICD Tools Plugin to UI

NOTE: Will not merge until the new year

## Description

Changes in 3 places:
- Added as option to "Inventory" dropdown
- Added as result in "Inventory" tab of Results page
- Added as a filter option on Repository tab of Search page
  - Creates a new dropdown component
    - Only "exact" matches work with key lookups and the exact matches would have to be with an un-intuitive ID we define. Because of this, a dropdown leads to the best user experience

### Inventory dropdown:
Screenshot 2024-12-18 at 10 49 01 AM

#### No results
Screenshot 2024-12-18 at 11 24 22 AM

#### Results
Screenshot 2024-12-18 at 10 48 50 AM

#### On first load
Screenshot 2024-12-18 at 10 43 23 AM

#### Dropdown
Screenshot 2024-12-18 at 10 43 31 AM

#### Selected
Screenshot 2024-12-18 at 10 43 42 AM

## Motivation and Context

- These features have been enabled API side for a bit, but we haven't exposed them as of yet

## How Has This Been Tested?

- Tests pass
- Working in dev environment

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/ui/src/features/search/searchSchemas.ts
+++ b/ui/src/features/search/searchSchemas.ts
@@ -191,7 +191,10 @@ export const supportedCicdTools = [
 	},
 ];
 
-const cicdToolIDs = supportedCicdTools.map((item) => item.id);
+const cicdToolIDs = [
+	"", // Add empty string so that no value is acceptable
+	...supportedCicdTools.map((item) => item.id),
+];
 
 function getCicdToolError() {
 	const validToolsStr = `[${cicdToolIDs.join("|")}]`;
--- a/ui/src/features/search/searchSchemas.ts
+++ b/ui/src/features/search/searchSchemas.ts
@@ -148,6 +148,62 @@ export const serviceSchema = () => {
 		);
 };
 
+export const supportedCicdTools = [
+	{
+		id: "aws_codebuild",
+		displayName: "AWS CodeBuild",
+	},
+	{
+		id: "azure_pipelines",
+		displayName: "Azure Pipelines",
+	},
+	{
+		id: "bitrise",
+		displayName: "BitRise",
+	},
+	{
+		id: "circleci",
+		displayName: "CircleCI",
+	},
+	{
+		id: "electron_forge",
+		displayName: "Electron Forge",
+	},
+	{
+		id: "github_actions",
+		displayName: "GitHub Actions",
+	},
+	{
+		id: "gitlab_ci",
+		displayName: "GitLab CI",
+	},
+	{
+		id: "jenkins",
+		displayName: "Jenkins",
+	},
+	{
+		id: "teamcity",
+		displayName: "TeamCity",
+	},
+	{
+		id: "travis_ci",
+		displayName: "Travis CI",
+	},
+];
+
+const cicdToolIDs = supportedCicdTools.map((item) => item.id);
+
+function getCicdToolError() {
+	const validToolsStr = `[${cicdToolIDs.join("|")}]`;
+	const intro = i18n._(t`Must be a supported cicd tool name:`);
+
+	return `${intro} ${validToolsStr}`;
+}
+
+export const cicdToolSchema = () => {
+	return Yup.string().trim().oneOf(cicdToolIDs, getCicdToolError());
+};
+
 export const riskSchema = Yup.string()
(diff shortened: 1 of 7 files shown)


ui/src/features/search/searchSchemas.ts
```diff
@@ -148,6 +148,65 @@ export const serviceSchema = () => {
 		);
 };
 
+export const supportedCicdTools = [
+	{
+		id: "aws_codebuild",
+		displayName: "AWS CodeBuild",
```
g-marconet, line 154: These are the names of services. I imagine that is consistent across languages. Should these be translatable?

ui/src/pages/ResultsPage.tsx
```diff
@@ -4776,6 +4776,27 @@ const InventoryTabContent = (props: {
...
+			});
+		}
+		return result;
+	};
+
+	const cicdToolsColumns: ColDef[] = [
+		{ field: "tool", headerName: i18n._(t`Tool`) },
+		{ field: "files", headerName: i18n._(t`Config Files`) },
+	];
+	const cicdToolsRows: RowDef[] = getCICDRows(", ");
```
ZoogieZork, line 4798: Looking at sample output, I'm not keen about using commas here.  It's visually difficult to scan.
image

We've got a few different options here:

1. (easy) Increase the visual separation between the filenames.  This could be inserting more padding between the filenames or making the filenames be links to the repository (so they show up as a separate color to the separator).
2. (easy?) Separate the filenames with newlines -- not sure how nice this will play with the table layout.
3. Mimic the style of the Base Images section with each (Tool, Filename) be on its own row
  image

My preference would be 2, then 1, then 3, but we might want to get other opinions as well.

ui/src/pages/ResultsPage.tsx
```diff
@@ -4916,19 +4937,45 @@ const InventoryTabContent = (props: {
...
+				</Toolbar>
+				{(scan.results_summary?.inventory?.cicd_tools || 0) > 0 ? (
+					<EnhancedTable
+						columns={cicdToolsColumns}
+						rows={cicdToolsRows}
+						defaultOrderBy="tool"
+						menuOptions={{
+							exportFile: "cicd_tools",
+							exportFormats: ["csv", "json"],
+							exportData: getCICDRows,
```
ZoogieZork, line 4972: I noticed the exported JSON has the filenames separated by commas instead of being an array -- anything we can do about that?

```json
{
  "tool": "Jenkins",
  "files": "Jenkinsfile,build/Jenkinsfile"
}
```

ui/src/features/search/searchSchemas.ts
ZoogieZork: These are all proper nouns, so I don't _think_ they would be translated.  Let's leave them untranslated -- we can add translations later if we find a counterexample.

ZoogieZork (changes requested):

ui/src/pages/ResultsPage.tsx
g-marconet: This was intended to clean output for CSVs, but it turns out we convert arrays to escaped comma-separated strings automatically. Good catch

ui/src/pages/ResultsPage.tsx
g-marconet: How about as a list?

Screenshot 2024-12-19 at 12 03 27 PM

- 2 as written didn't look right to me. The bullets make that a little better, IMO 
- 3 is unintuitive to me. The files are more justification for the assertion that the repository uses the given CI/CD Tool than a new finding themselves (like a new Tag is for the base images plugin)
- I agree that 1 with links to the files is ideal, but the CICD Tools plugin does not currently return a URL to the file. We could theoretically derive links from within the UI, but that'd be a lot of VCS-specific logic in the UI and it's unclear we'd have everything we need.

ZoogieZork: Yeah, the bullets look great -- I think we've got a winner here.

ZoogieZork (approved):

PR #403, opened 2025-01-14
Engine - Gracefully shutdown containers when timeout is exceeded

g-marconet:
Stops the docker container when timeout is reached.

## Description

- Runs `docker stop <container-name>` when plugin timeout has been reached
  - Previously, we just let `subprocess.run` kill the original `docker run ...` process, but the container would still run in the background. With this change, we tell the container to stop with SIGTERM, give it the default 10s grace period, then SIGKILL if it still hasn't exited.
- As a note, this currently doesn't do anything, as no active plugins specify a timeout in their `settings.json` and we do not set a default timeout.
  - As a result, this change was tested by hardcoding the timeout to a small value.

## Motivation and Context

- Containers currently continue running, orphaned, after the timeout has been reached. This fixes that.

## How Has This Been Tested?

- Tested in dev environment

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/backend/engine/utils/plugin.py
+++ b/backend/engine/utils/plugin.py
@@ -375,8 +375,9 @@ def run_plugin(
     # The temporary named volume is automatically deleted after the plugin
     # container exits.
     with temporary_volume(f"{TEMP_VOLUME_NAME_PREFIX}-{plugin}") as volname:
+        container_name = get_container_name()
         plugin_command = get_plugin_command(
-            scan, plugin, settings, depth, include_dev, volname, scan_images, plugin_config, services
+            scan, plugin, container_name, settings, depth, include_dev, volname, scan_images, plugin_config, services
         )
         try:
             # Run the plugin inside the settings.image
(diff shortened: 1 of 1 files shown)


mdfleury-wbd (approved):

backend/engine/utils/plugin.py
```diff
@@ -388,6 +389,20 @@ def run_plugin(
...
             )
         except subprocess.TimeoutExpired:
+            stop_command = get_plugin_stop_command(container_name)
+            stop_response = subprocess.run(
+                stop_command,
+                stdout=subprocess.PIPE,
+                stderr=subprocess.PIPE,
+                encoding="utf-8",
+                check=False,
+            )
```
ZoogieZork, line 399: Moving forward, we should use the Docker SDK for Python rather than calling out to the Docker CLI.  Besides being more efficient, this will remove any potential confusion between if an error is coming from the CLI or the Docker service itself (since the Docker SDK talks to the service socket directly) and will allow us to easily create mocks for unit tests.

That said, fixing the potential "hanging container" issue is more important right now, so this is not a blocker.  We already have a separate backlog task to migrate to the Docker SDK (currently focused on the plugins right now), so we'll migrate this later down the road.

ZoogieZork (approved):

PR #417, opened 2025-02-05
Fix `is_valid_timestamp` for strings with timezone

g-marconet:
## Description
- Timestamp format was `%Y-%m-%dT%H:%M:%SZ`
  - This works fine with times in UTC (with a "Z" at the end), but doesn't parse non-UTC timezones
- Gitlab seems to not use UTC. This example was taken from their docs: `2024-06-28T03:44:20-07:00`
  - This would fail with the previous timestamp format

## Motivation and Context

Previously, this bug would cause issues retrieving timestamp information for Gitlab.

## How Has This Been Tested?

Tested in dev environment and by crafting strings

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/orchestrator/lambdas/layers/heimdall_utils/heimdall_utils/utils.py
+++ b/orchestrator/lambdas/layers/heimdall_utils/heimdall_utils/utils.py
@@ -3,7 +3,7 @@ import json
 import logging
 import time
 from collections import namedtuple
-from datetime import datetime, timedelta
+from datetime import datetime, timedelta, timezone
 from json import JSONDecodeError
 from typing import Optional
 
@@ -133,6 +133,6 @@ def is_valid_timestamp(timestamp: str) -> bool:
 
 
 def get_datetime_from_past(days: int) -> datetime:
-    current_timestamp = datetime.now()
+    current_timestamp = datetime.now(timezone.utc)
     three_months_ago = current_timestamp - timedelta(days=days)
     return three_months_ago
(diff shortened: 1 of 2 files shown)


orchestrator/lambdas/layers/heimdall_utils/heimdall_utils/utils.py
```diff
@@ -17,7 +17,7 @@
 logging.getLogger("botocore").setLevel(logging.WARNING)
 
 DYNAMODB_TTL_DAYS = 60
-TIMESTAMP_FORMAT = "%Y-%m-%dT%H:%M:%SZ"
+TIMESTAMP_FORMAT = "%Y-%m-%dT%H:%M:%S%z"
```
ZoogieZork, line 20: This changes the format of the default timestamp returned by `parse_timestamp()`.

Before: `2025-02-06T00:22:45Z`
After: `2025-02-06T00:22:45+0000`

This should be fine -- we just need to update the documentation for `parse_timestamp()` to reflect the change.

ZoogieZork (changes requested):
Just need a documentation update.

orchestrator/lambdas/layers/heimdall_utils/heimdall_utils/utils.py
g-marconet: Good catch. FWIW, I believe this only effects the cases where we generate a new Timestamp. Valid UTC timestamps with a `Z` instead of `+0000` will still maintain the Z (for example, `parse_timestamp("2025-02-06T00:22:45Z")` will output `"2025-02-06T00:22:45Z"` since it just returns the original timestamp.)

ZoogieZork: We should also delete this line in the doc:

```
- All returned timestamps are in UTC (denoted by the 'Z' suffix).
```

ZoogieZork (approved):

PR #467, opened 2025-05-05
Handle exception in users lambdas

g-marconet:
Handles `JSONDecodeError`s in the users lambdas

## Description

- `users_keys` and `users_services` would occasionally raise a `JSONDecodeError` with message `Expecting value: line 1 column 1 (char 0)`. We can reproduce this error message outside of Artemis with `python -c "import json; json.loads('')"`, implying the problem is when the body is empty string.
- The existing `or "{}"` should cover this case, as empty string is falsey. So, something else is going. That said, any `JSONDecodeError` should be treated as validation failing, so we handle that case generically by raising `ValidationError` instead of handling this particular error.

## Motivation and Context

- These lambdas were occasionally crashing with errors. This PR handles the underlying error that caused these crashes.

## How Has This Been Tested?

Working in dev environment

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/backend/lambdas/api/users_keys/users_keys/util/parsers.py
+++ b/backend/lambdas/api/users_keys/users_keys/util/parsers.py
@@ -42,7 +42,10 @@ def parse_event(event):
 
 
 def parse_body(event, admin, features, user_id):
-    body = json.loads(event.get("body") or "{}")
+    try:
+        body = json.loads(event.get("body") or "{}")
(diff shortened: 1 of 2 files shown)


ZoogieZork (approved):
Looks like the body is passed directly from the HTTP request, so catching the `JSONDecodeError` is the right thing to do, regardless of what's been triggering this.

PR #471, opened 2025-05-12
Generate composer.lock files

faizanH:
## Description

Currently, when a directory has a composer.json file and does not have a composer.lock file, we can not use our trivy SCA scanner. Now with this change, a composer.lock file will be auto generated when a directory contains a composer.json file.
## Motivation and Context

To extend our SCA language coverage
## How Has This Been Tested?

in a development environment
## Types of changes

- [ ] Documentation change

## Checklist


backend/engine/plugins/lib/trivy_common/generate_composer_locks.py
```diff
@@ -0,0 +1,109 @@
...
+    "composer.lock"
+]
+
+docker_client = docker.from_env()
+
+
+def install_package_files(include_dev: bool, path: str, root_path: str):
+    # Create a composer.lock file if it doesn't already exist
+    logger.info(
+        f"Generating composer.lock for {path.replace(root_path, '')} (including dev dependencies: {include_dev}"
```
ZoogieZork, line 27: (minor) Missing end paren at the end of the log message.

backend/engine/plugins/trivy_sca/main.py
```diff
@@ -54,11 +64,7 @@ def main():
         results.extend(result)
 
     # Return results
-    print(
-        json.dumps(
-            {"success": not bool(results), "details": results, "errors": lock_file_errors, "alerts": lock_file_alerts}
-        )
-    )
+    print(json.dumps({"success": not bool(results), "details": results, "errors": errors, "alerts": alerts}))
```
ZoogieZork, line 67: (minor) Since our JSON results can be quite large, we should switch from using `json.dumps` (which renders the whole JSON to a string first) to `json.dump` that renders the output directly to the output, e.g.:

```python
json.dump({ ... }, sys.stdout)
```

backend/engine/plugins/lib/trivy_common/generate_composer_locks.py
```diff
@@ -0,0 +1,109 @@
...
+    )
+    if not include_dev:
+        cmd.append("--no-dev")
+
+    # Run Composer in a container
+    COMPOSER_IMG = "composer:latest"
+    container_name = "composer_runner"
```
ZoogieZork, line 34: Container names must be unique, so we should append a UUID.

PR #482, opened 2025-06-20
Catch OS errors from broken symlinks

mdfleury-wbd:
## Types of changes

- [ ] Documentation change

## Checklist

## Pic
Embed something funny here

### DIFF
--- a/backend/engine/plugins/cicd_tools/detectors/aws_codebuild_detector.py
+++ b/backend/engine/plugins/cicd_tools/detectors/aws_codebuild_detector.py
@@ -1,5 +1,3 @@
-import json
-
 from pathlib import Path
 from typing import Callable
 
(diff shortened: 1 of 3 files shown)


backend/engine/plugins/cicd_tools/detectors/pattern_detector.py
```diff
@@ -53,6 +53,8 @@ def _get_matches(self, path: str) -> list[Path]:
         base = Path(path)
 
         for pattern in self.patterns:
-            result.extend(base.rglob(pattern))
-
+            try:
+                result.extend(base.rglob(pattern))
+            except OSError:
+                continue
```
g-marconet, line 59: Why don't we append to the `errors` list here?

backend/engine/plugins/cicd_tools/detectors/aws_codebuild_detector.py
```diff
@@ -33,22 +31,24 @@ def check(self, path: str) -> DetectorResult:
...
                 result["in_use"] = True
-                result["configs"].append({"path": str(file.relative_to(path))})
-
-        for config in buildspec_files:
-            result["in_use"] = True
-            result["configs"].append({"path": str(config.relative_to(path))})
+                result["configs"].append({"path": str(config.relative_to(path))})
 
+        except OSError as e:
+            result["errors"].append(f"Error during CodeBuild detection: {e}")
```
g-marconet, line 53: Would these be better as `alerts`?

I don't think we have any guidelines for when to use one over the other. In my head, a broken symlink would also break a CICD tool that follows the link looking for a config. As a result, we likely aren't missing any valid config files when we see an error from a broken symlink and so this might be more of an `alert` than an `error`. What do you think?

backend/engine/plugins/cicd_tools/detectors/pattern_detector.py
g-marconet: I meant `PatternDetector` in general. We could, for example, modify this function to return a tuple including a list of strings to add to errors/alerts.

g-marconet (approved):
LGTM

PR #505, opened 2025-07-24
Improve error checking in bandit plugin

mdfleury-wbd:
## How Has This Been Tested?
Tested in nonprod

## Types of changes

- [ ] Documentation change

## Checklist

## Pic
Embed something funny here

### DIFF
--- a/backend/engine/plugins/bandit/settings.json
+++ b/backend/engine/plugins/bandit/settings.json
@@ -1,5 +1,6 @@
 {
-    "name": "Bandit",
-    "type": "static_analysis",
-    "image": "$ECR/artemis/python:3-latest"
+  "name": "Bandit",
+  "type": "static_analysis",
+  "image": "$ECR/artemis/python:3-latest",
+  "writable": true
 }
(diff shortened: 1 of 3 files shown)


backend/engine/plugins/bandit/settings.json
```diff
@@ -1,5 +1,6 @@
 {
-    "name": "Bandit",
-    "type": "static_analysis",
-    "image": "$ECR/artemis/python:3-latest"
-}
\ No newline at end of file
+  "name": "Bandit",
+  "type": "static_analysis",
+  "image": "$ECR/artemis/python:3-latest",
+  "writable": true
```
ZoogieZork, line 5: We want to avoid enabling `writable` whenever possible, since it will prevent the plugin from being run in parallel (in the future).

`tempfile.TemporaryDirectory` _should_ create the temp dir outside of the target directory (e.g. `/tmp`).  If not, we may need to specify the `dir` arg.

If Bandit itself is trying to write to the target (e.g. the sources being scanned), then maybe there's a way around it (like changing the working directory?).

ZoogieZork (changes requested):

PR #522, opened 2025-08-05
Add lint checks for UI unit tests

ZoogieZork:
## Description

Adds lint checks for the UI unit tests, via [eslint-plugin-jest](https://www.npmjs.com/package/eslint-plugin-jest).

Both recommended and style rules are enabled.

The jest/no-conditional-expect rule is disabled for now since we use conditional assertions extensively.  We can examine and resolve these individual cases later.

## Motivation and Context

Identify hidden issues that would result in failures in Jest v30.

## How Has This Been Tested?

Run all unit tests.

## Types of changes

- [ ] Documentation change

## Checklist

### DIFF
--- a/ui/eslint.config.mjs
+++ b/ui/eslint.config.mjs
@@ -410,6 +410,7 @@ export default defineConfig([
 			...jest.configs.recommended.rules,
 			...jest.configs.style.rules,
 			"jest/no-conditional-expect": "off", // Needs additional review.
+			"jest/no-disabled-tests": "off", // We purposefully skip some flaky tests pending investigation.
 			"jest/expect-expect": [
 				"warn",
 				{


ui/src/components/__tests__/MailToLink.test.tsx
```diff
@@ -127,14 +127,6 @@ describe("MailToLink component", () => {
 		expect(link).toHaveAttribute("href", `mailto:${recipient}`);
 	});
 
-	test("uses supplied recipient if valid email address", () => {
```
g-marconet, line 130: Why is this test being removed?

ZoogieZork: This test is an exact duplicate of the one above, probably a copy-paste mishap.

PR #532, opened 2025-08-13
Add key reminder lambda

mdfleury-wbd:
Adding new scheduled lambda to remind users that they have keys that are going to expire soon

## How Has This Been Tested?
Tested in nonprod

## Types of changes

- [ ] Documentation change

## Checklist

## Pic
Embed something funny here


backend/lambdas/scheduled/key_reminder/key_reminder/handlers.py
```diff
@@ -0,0 +1,99 @@
...
+    expired = []
+
+    for key in api_keys:
+        expires = key.expires
+        if expires is None:
+            continue
+        delta = expires - now
+        if delta.days == 30:
+            in_30_days.append((key))
+        if delta.days == 7:
```
breedenc, line 32: Would it make sense to make these either `elif`s or match/case structure?

backend/lambdas/scheduled/key_reminder/key_reminder/handlers.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def notify_user(days: int, keys):
+    message = (
+        'Your key "{name}" will expire {expiration}.\nYou can see a list of your keys here: https://{domain}/settings'
+    )
+    for key in keys:
+        name = key.name
+        expiration = ""
+        match days:
```
breedenc, line 62: This doesn't make a change to functionality, but it might be clearer reading the code if the match/case is pulled out of the loop since `days` is set when calling `notify_user`?

backend/pytest.ini
```diff
@@ -32,3 +33,5 @@ filterwarnings =
     # Deprecation warning from using boto3, ignore until it's fixed in botocore:
     # https://github.com/boto/boto3/issues/3889
     ignore:datetime.datetime.utcnow\(\) is deprecated:DeprecationWarning:botocore.*
+; change this to 1 to see logs
+log_cli = 0
```
breedenc, line 37: - Is this change related to the feature at all or just a preference thing?
- Is it possible to configure this with command-line args for pytest (and maybe additional `make` targets) so that the developer can specify the log level desired when running the tests rather than it being set here?

